PR #1, opened 2020-03-28
Add models

AlexVPopov:
* Add the models, as per https://dbdiagram.io/d/5e7e3b434495b02c3b88df64
* add faker for generating fake data (will be useful for the seed script);


app/models/user.rb
```diff
@@ -0,0 +1,15 @@
+class User < ApplicationRecord
+  belongs_to :organisation
+  has_many :contact_list_users
+  has_many :contact_lists, through: :contact_list_users
+  has_many :users, through: :contact_lists
```
jamiebuckley: Does a user have many users?  I would have thought if there was a relationship here it would have been the list of contacts from all the lists the user is assigned to

jamiebuckley: My rails is terrible, but something like:
`has_many :contacts, through: :contact_lists`

AlexVPopov: Absolutely my bad, of course this is a mistake, fixing now :)

PR #4, opened 2020-03-28
Feature/38 add user invite functionality rebase

jamiebuckley:


config/routes.rb
```diff
@@ -1,5 +1,6 @@
 Rails.application.routes.draw do
   resources :contacts
+  resources :users
```
AlexVPopov: I'd scope this to:

```suggestion
  resources :users, only: [:index, :new, :create]
```

db/migrate/20200328103438_add_attributes_to_users.rb
```diff
@@ -0,0 +1,12 @@
+class AddAttributesToUsers < ActiveRecord::Migration[6.0]
+  def change
+    add_column :users, :email, :text, null: false
```
AlexVPopov: I'd suggest a unique index on the `email`column.

app/models/user.rb
```diff
@@ -1,5 +1,5 @@
 class User < ApplicationRecord
-  belongs_to :organisation
+  belongs_to :organisation, optional: true
```
AlexVPopov: Also, please add these changes (optional belongs_to, removed validations) to the user specs as well.

db/migrate/20200328103438_add_attributes_to_users.rb
```diff
@@ -1,6 +1,7 @@
 class AddAttributesToUsers < ActiveRecord::Migration[6.0]
   def change
     add_column :users, :email, :text, null: false
+    add_index :users, :email
```
AlexVPopov: Perfect, I'd only add `unique: true` as well.

spec/factories/users.rb
```diff
@@ -4,5 +4,7 @@
 
     first_name { Faker::Name.first_name }
     last_name { Faker::Name.last_name }
+    email { Faker::Internet.email }
```
AlexVPopov: You can add uniqueness here as well: `Faker::Internet.unique.email` (from https://github.com/faker-ruby/faker#ensuring-unique-values).

db/migrate/20200328103438_add_attributes_to_users.rb
jamiebuckley: no probs, will add that too

PR #9, opened 2020-03-28
Use omniauth-auth0

timwis:
Do not merge until we decide which library to use. This is just one option.

Based on existing omniauth-onetime branch and [Auth0's official tutorial](https://auth0.com/docs/quickstart/webapp/rails).


timwis:
I've just added a commit that keeps Auth0 in sync with our users table. It works without this: prior to this commit, if an auth0 user didn't exist in our users table it would fail to log them in. But auth0 would create a user in their database for everyone who attempts to sign in. This commit allows us to disable auth0 sign ups: instead, you add users in our application and the application creates the user in the auth0 database via the auth0 management API. When you delete a user in our application, it deletes it in the auth0 database as well. If the auth0 requests fail, the application doesn't create/delete the user in its database either and instead shows an error.

Configuring auth0 is a bit of a pain. I've added docs at the end of the readme. Apologies for formatting issues but that should be everything.

timwis:
Closing in favour of #10

PR #10, opened 2020-03-28
Use mikker/passwordless

timwis:
~~Do not merge until we decide which library to use. This is just one option.~~ Ready to review and merge.

Implements [mikker/passwordless](https://github.com/mikker/passwordless).


timwis:
Note that `schema.rb` includes a change to the `notes` table. That's just because I ran `rails db:drop`, `rails db:create`, and `rails db:migrate` after merging the latest master. I think the previous work on notes didn't include that merge is all, but it looks correct according to the latest migrations.

app/controllers/application_controller.rb
```diff
@@ -1,3 +1,17 @@
 class ApplicationController < ActionController::Base
+  include Passwordless::ControllerHelpers
   http_basic_authenticate_with name: 'camden', password: 'camden'
```
eddieleeper-nb: `http_basic_authentication` should be removed

app/controllers/users_controller.rb
```diff
@@ -1,6 +1,7 @@
 class UsersController < ApplicationController
 
   before_action :set_user, only: [:destroy]
+  #before_action :require_user!
```
eddieleeper-nb: Should this be enabled really?

config/environments/development.rb
```diff
@@ -36,6 +36,8 @@
 
   config.action_mailer.perform_caching = false
 
+  config.action_mailer.default_url_options = { host: 'localhost:3000' }
+
```
eddieleeper-nb: This doesn't look right, and I think could be removed if this branch is updated with the latest master branch.  Jamie has added Mailer configuration options.

config/initializers/passwordless.rb
```diff
@@ -0,0 +1,4 @@
+Passwordless.default_from_address = "CHANGE_ME@example.com"
```
eddieleeper-nb: We should probably ask the customer for an address to use here.

config/routes.rb
```diff
@@ -1,6 +1,7 @@
 Rails.application.routes.draw do
   resources :contacts, only: [:index]
   resources :users, only: [:index, :new, :create, :destroy]
+  passwordless_for :users, at: '/', as: :auth
```
eddieleeper-nb: We are going to need to secure all routes I guess, but we will also require a way to login for the first users.  A database seed script with key email addresses for users in production.

db/schema.rb
```diff
@@ -52,7 +52,7 @@
   end
 
   create_table "notes", force: :cascade do |t|
-    t.text "body"
+    t.text "body", null: false
```
eddieleeper-nb: There might be a conflict here from latest master - my fault I missed this from a PR, but Jamie added it.

app/controllers/application_controller.rb
eddieleeper-nb: Actually - depends whose PR gets merged first - Alex has removed it in his PR.

app/controllers/users_controller.rb
timwis: I expect we want to enable it so that only logged in users can view this controller. I left it commented for the time being because the other controller was broken, so this one was the only way to test.

config/environments/development.rb
timwis: This actually sets that hostname to use in the link that gets emailed to the user, e.g. `http://localhost:3000/signin?token=xyz`. I agree it felt strange to put there, but it's in the passwordless docs and fails if it's not set. Perhaps @jamiebuckley 's updates supercede this.

config/initializers/passwordless.rb
timwis: Captured in trello task

config/routes.rb
timwis: Captured in trello task

config/environments/development.rb
timwis: Captured in trello task

PR #13, opened 2020-03-29
Add notes / comment model

eddieleeper-nb:
Added notes model

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -1,6 +1,6 @@
 class Contact < ApplicationRecord
   belongs_to :contact_list, optional: true
-  has_many :notes
+  has_many :notes, dependent: :destroy
   has_many :tasks, dependent: :destroy
   has_many :uncompleted_tasks, -> { uncompleted }, class_name: 'Task'
   has_many :completed_tasks, -> { completed }, class_name: 'Task'
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -1,5 +1,6 @@
 class Contact < ApplicationRecord
   belongs_to :contact_list, optional: true
+  has_many :notes
   has_many :tasks, dependent: :destroy
   has_many :uncompleted_tasks, -> { uncompleted }, class_name: 'Task'
   has_many :completed_tasks, -> { completed }, class_name: 'Task'
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -1,0 +1,3 @@
+class Note < ApplicationRecord
+  belongs_to :contact
+end
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -1,3 +1,5 @@
 class Note < ApplicationRecord
   belongs_to :contact
+
+  validates :body, presence: true
 end
--- a/db/migrate/20200329100626_create_notes.rb
+++ b/db/migrate/20200329100626_create_notes.rb
@@ -1,7 +1,7 @@
 class CreateNotes < ActiveRecord::Migration[6.0]
   def change
     create_table :notes do |t|
-      t.text :note
+      t.text :body, null: false
       t.references :contact, null: false, foreign_key: true
 
       t.timestamps
--- a/db/migrate/20200329100626_create_notes.rb
+++ b/db/migrate/20200329100626_create_notes.rb
@@ -1,0 +1,10 @@
+class CreateNotes < ActiveRecord::Migration[6.0]
+  def change
+    create_table :notes do |t|
+      t.text :note
+      t.references :contact, null: false, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
--- a/spec/factories/notes.rb
+++ b/spec/factories/notes.rb
@@ -1,0 +1,6 @@
+FactoryBot.define do
+  factory :note do
+    note { "MyText" }
+    contact { nil }
+  end
+end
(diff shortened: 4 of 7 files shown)


app/models/contact.rb
```diff
@@ -1,5 +1,6 @@
 class Contact < ApplicationRecord
   belongs_to :contact_list, optional: true
+  has_many :notes
```
AlexVPopov: I'd add `dependent: :destroy`.

db/migrate/20200329100626_create_notes.rb
```diff
@@ -0,0 +1,10 @@
+class CreateNotes < ActiveRecord::Migration[6.0]
+  def change
+    create_table :notes do |t|
+      t.text :note
```
AlexVPopov: I'd add `null: fase`.

AlexVPopov: I think a name like `body` or `content` or something similar, would be more suitable, as otherwise we'd have `note.note`.

app/models/note.rb
```diff
@@ -0,0 +1,3 @@
+class Note < ApplicationRecord
+  belongs_to :contact
```
AlexVPopov: I'd add a validation on the note/body.

spec/factories/notes.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :note do
+    note { "MyText" }
```
AlexVPopov: You can use `Faker::Lorem` for the contents: https://github.com/faker-ruby/faker/blob/master/doc/default/lorem.md

spec/factories/notes.rb
```diff
@@ -0,0 +1,6 @@
...
+    contact { nil }
```
jamiebuckley: Is it possible to reference the Contact factory here, just thinking that contact is a not-null field?  Don't know whether it matters on not, I'm new to this FactoryBot stuff

AlexVPopov: I think it'd be better, if factory bot generates a note with a contact by default (as `belongs_to` is required by default). For this to happen, we need to put `contact` at the beginning: https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#associations

spec/models/note_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Note, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
AlexVPopov: I'd add basic specs for associations and validations, as similar to the other models.

spec/models/note_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Note, type: :model do
```
jamiebuckley: Worth deleting this or filling it out?

spec/factories/notes.rb
eddieleeper-nb: @AlexVPopov Can you help answer Jamies question - I'm not an expert on factory bot either.

spec/models/note_spec.rb
eddieleeper-nb: It 'should' get filled out really, but I can delete for MVP. what approach have you guys been doing yesterday?  Did you talk tests? I didn't see anything in the main chat.

spec/factories/notes.rb
AlexVPopov: Yes, the answe is this:

```ruby 
FactoryBot.define do
  factory :note do
    contact

    note { Faker::Lorem.sentence }
  end
end
```

spec/models/note_spec.rb
AlexVPopov: Yes, we've added some basic tests, check out the other model specs.

jamiebuckley: I've been writing the specs.  I did start on doing a proper integration test for a controller, but was concerned by amount of work vs payoff it would be given the current changes in designs.

Agree we need to have a chat about test strategy soon though

db/migrate/20200329100626_create_notes.rb
eddieleeper-nb: I was going from the database design

AlexVPopov: Yeah, sorry, I agree.

app/models/contact.rb
eddieleeper-nb: Added.

spec/models/note_spec.rb
AlexVPopov: Agree. My ad-hoc strategy so far - write quick unit-tests only, skip integration tests for now. The unit tests just guard against small overlooks like typos, breaking associations and stuff. I don't think we need intricate tests for the MVP.

db/migrate/20200329100626_create_notes.rb
eddieleeper-nb: I've updated as per your suggestions @AlexVPopov

app/models/note.rb
eddieleeper-nb: Done

spec/factories/notes.rb
eddieleeper-nb: thanks - I've updated.

PR #14, opened 2020-03-29
Feature/48 implement people update flow

jamiebuckley:
Screenshot from 2020-03-29 13-27-42

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_03_28_201840) do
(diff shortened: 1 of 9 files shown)


db/schema.rb
```diff
@@ -47,12 +47,13 @@
...
     t.integer "uncompleted_tasks_count", default: 0
     t.integer "completed_tasks_count", default: 0
+    t.text "email"
     t.index ["contact_list_id"], name: "index_contacts_on_contact_list_id"
     t.index ["priority"], name: "index_contacts_on_priority"
   end
 
   create_table "notes", force: :cascade do |t|
-    t.text "body"
+    t.text "body", null: false
```
AlexVPopov: I think this change has appeared due to a previous migration?

PR #15, opened 2020-03-29
Implemented notes for contacts.

eddieleeper-nb:


config/routes.rb
```diff
@@ -1,5 +1,9 @@
 Rails.application.routes.draw do
-  resources :contacts, only: [:index, :edit, :update]
+
+  resources :contacts, only: [:index, :show, :edit, :update] do
+    resources :notes
```
AlexVPopov: I'd scope this to only:
```suggestion
    resources :notes, only: [:create]
```

PR #16, opened 2020-03-29
Tasks index

AlexVPopov:
- add tasks and my-tasks views;
- add pagiantion to tasks, contacts, users;
- add more seed data;

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -14,4 +14,7 @@ class Task < ApplicationRecord
 
 
   validates :name, presence: true
+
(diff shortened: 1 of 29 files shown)


app/models/task.rb
```diff
@@ -14,4 +14,7 @@ class Task < ApplicationRecord
 
 
   validates :name, presence: true
+
+  delegate :name, :priority, to: :contact, prefix: true
+  delegate :name, to: :user, prefix: true
```
jamiebuckley: This is a really nice little syntax feature

PR #17, opened 2020-03-30
Feature/18 add tasks to individual contact view

jamiebuckley:
Screenshot from 2020-03-30 16-36-23
Screenshot from 2020-03-30 16-39-16


app/views/application/_tasks_table.html.erb
```diff
@@ -13,7 +13,11 @@
   <tbody>
     <% tasks.each do |task| %>
     <tr>
-      <td><%= task.user_name %></td>
+      <% if task.user %>
+        <td>Assigned to <%= task.user_name %></td>
+      <% else %>
+        <td>No user</td>
```
eddieleeper-nb: In the drop down when creating a task the text is `Unassigned` should use the same word here.

app/views/contacts/show.html.erb
```diff
@@ -2,53 +2,56 @@
 <p class="notice"><%= notice %></p>
 <% end %>
 
-<%= link_to contacts_path, class: "go-back" do %>
-  <%= image_tag "back.svg", alt: "" %><span>Go back</span>
-<% end %>
+<div class="tab_buttons margin-bottom-70">
+  <%= link_to 'Profile', contact_path(@contact), class: 'active' %>
+  <%= link_to "View #{@contact.first_name}'s Tasks (#{@contact.tasks.count})", controller: :contacts, action: :show_tasks %>
```
eddieleeper-nb: Not sure the wording here makes sense - TBF - I couldn't see any wording on the designs.
Instead of "`FirstName`'s Tasks" which sounds like its the contact who needs to do something, how about:
`View tasks to do for first_name`
Maybe check with Nicholas/Jaye/Tim - not too important, but words matter :-)

config/routes.rb
```diff
@@ -1,8 +1,10 @@
 Rails.application.routes.draw do
   resources :contacts, only: [:index, :show, :edit, :update] do
-    resources :notes, only: [:create]
+    resources :tasks, only: [:create]
   end
-  resources :tasks, only: [:index]
+  get '/contacts/:id/tasks', to: 'contacts#show_tasks'
```
eddieleeper-nb: I think that there is a better way to do this - but it's good for now - it works and I don't have the answer.

eddieleeper-nb: I think there is a better way to do this, but I don't have the answer at the moment, so it's good.

PR #18, opened 2020-03-30
Feature/styled login

jhackett1:


config/database.yml
```diff
@@ -8,14 +8,14 @@ default: &default
 development:
   <<: *default
   url: postgresql://localhost/i_have_i_need_development
-  username: dev_local_user
-  password: password
+  # username: dev_local_user
```
timwis: Hm, what's this for? Seems unrelated to this feature, no?

app/views/passwordless/sessions/create.html.erb
```diff
@@ -0,0 +1,9 @@
+<% content_for :class, "login-panel--centred"%>
+
+<%= image_tag "green-tick.svg", alt: "success" %>
+
+<h2>Check your email</h2>
+
+<p>If <strong>firstname.lastname@gmail.com</strong> exists in our records, you'll get an email with a special link.</p>
```
timwis: Did you mean to use the email address the user entered here, rather than a hardcoded gmail address? Also, while we're at it, we may want to say, "If you don't get an email, speak to _xyz_" or that "you may not have an account setup by an administrator yet."

app/controllers/passwordless/sessions_controller.rb
```diff
@@ -0,0 +1,127 @@
+# frozen_string_literal: true
+
+require "bcrypt"
+
+module Passwordless
+  # Controller for managing Passwordless sessions
+  class SessionsController < ApplicationController
```
timwis: Why has this file been added? I was under the impression it's imported automatically by the gem and we can keep it out of the application code?

app/controllers/passwordless/application_controller.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Passwordless
+    # Base for Passwordless controllers
+    class ApplicationController < ::ApplicationController
```
timwis: Same question as below - why does this bit need to be in our application code? The Passwordless should add this behind the scenes.

app/controllers/passwordless/sessions_controller.rb
jhackett1: we need to modify the render call slightly to get it to call the right layout

config/database.yml
jhackett1: lol, this is me making it work with my local postgres. undone

app/views/passwordless/sessions/create.html.erb
jhackett1: i did some more looking and figured out how to fish out the email and show it on this screen

PR #27, opened 2020-03-31
Require login for secure routes

jamiebuckley:


jamiebuckley:
atm SessionsController also inherits ApplicationController, which makes it tricky

PR #28, opened 2020-03-31
73-add-notes-tasks views

eddieleeper-nb:


yarn.lock
```diff
@@ -2286,6 +2286,21 @@ dashdash@^1.12.0:
   dependencies:
     assert-plus "^1.0.0"
 
+datatables.net-dt@^1.10.20:
```
jamiebuckley: Note for future reviewer - this file is here because I missed it in my PR

rubyforbusiness: NB: had to do this to get the rails server working:
`yarn install --check-files`

This warning was related but I don't know enough to worry about it:
`warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.`

rubyforbusiness:
It's possible some manual intervention might be needed on heroku re the yarn update. We'll see.

PR #30, opened 2020-03-31
96 add user to note

eddieleeper-nb:


db/migrate/20200331160610_add_user_to_note.rb
```diff
@@ -0,0 +1,7 @@
+class AddUserToNote < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :notes, :user, null: false, foreign_key: true, index: {algorithm: :concurrently}
```
rubyforbusiness: This causes migrations to fail if they already have notes records. We should explicitly remove all the notes records just prior to this. I'll add a change to that effect.

PR #44, opened 2020-04-01
Ensure categories are inserted in lower case

jamiebuckley:
Search is case sensitive.  Categories are converted to snake_case and then humanized again, so 'Staying Social' becomes 'Staying social'.
This makes sure that categories are inserted into the db in lower case, but keeps the display of the categories as is

### DIFF
--- a/app/controllers/needs_controller.rb
+++ b/app/controllers/needs_controller.rb
@@ -34,7 +34,7 @@ class NeedsController < ApplicationController
   def create
     needs_params["needs_list"].each do |_, value|
       if value["active"] == "true"
-        need_category = value["name"].humanize
+        need_category = value["name"].humanize.downcase
         need_description = value["description"]
         if need_description.blank?
           need_description = "#{@contact.name} needs #{need_category}"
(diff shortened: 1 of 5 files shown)


timwis:
Maybe I don't fully understand the bug this fixes, but I would have assumed that, assuming the category is stored in the database as the exact name of the category, e.g. `Physical and mental wellbeing`, then I'd expect the navbar links to convert it to a slug (`physical-and-mental-wellbeing`) or snake (`physical_and_mental_wellbeing`) and then the filtering logic compares the query param to some kind of virtual field on the model. If we weren't using an ORM I'd assume the SQL `WHERE` clause would convert the category value to the slug/snake, e.g. `WHERE REPLACE(' ', '_', DOWNCASE(category)) = $1`.

I don't mean to block it; I'm sure you've thought through this, just throwing that out there.

jamiebuckley:
@timwis That's one way of doing it.  Essentially Rails doesn't seem to have an sql injection proof way of doing things like DOWNCASE, so the purpose of this is to store the category values downcased in the database, so that the query can use Ruby's `downcase` followed by ActiveRecord's `where`

jamiebuckley:
Basically it's simpler to keep it in the database in a consistent format rather than adding custom queries in the timeframe we have

timwis:
Got it. I assume the `humanize` method is basically 'to title case'. Code looks fine and simple. If you've QAd it, I think it's fine to merge 👍 unless you want deeper eyes on it.

jamiebuckley:
That's fine thanks, I'll do further testing on it in stage

PR #51, opened 2020-04-02
add reset filters link

rik86:
image


app/assets/stylesheets/task.scss
```diff
@@ -22,4 +22,9 @@
 
 .need__created{
     background: #F7C745;
+}
+
+a.reset-needs-filters{
```
timwis, line 27: Can we move this to [inside here](https://github.com/wearefuturegov/beacon/blob/master/app/assets/stylesheets/data_tables.scss#L117)? `task.scss` appears to be styles for viewing a need; the styles for the filters control on the needs list are in `data_tables.scss`.

rik86future: done

app/views/needs/index.html.erb
```diff
@@ -1,11 +1,16 @@
...
-  <h1 class="page-title">Needs <span class="page-title__count">(<%= @needs.total_count %>)</span></h1>
+  <h1 class="page-title">Needs</h1>
 </div>
 
-     
-
+<% if @needs.total_count > 0 %>
+  <div class="filtered-notice notice">
+    Showing <%= pluralize(@needs.total_count, 'result') %>
+    for the current filters.
```
timwis, line 10: Wouldn't this show even if no filters are set? And the "Clear filters" link would always be there too, right?

PR #56, opened 2020-04-02
silence any logs on the sign_in route including email content sent fr…

rik86:
…om Passwordless module

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -1,9 +1,2 @@
-# Be sure to restart your server when you modify this file.
-require 'silencer/logger'
-
 # Configure sensitive parameters which will be filtered from the log file.
 Rails.application.config.filter_parameters += [:password, :authenticity_token, :token, :email]
-Rails.application.config.action_mailer.logger = nil
-
-# This will silence all routing logs by replacing Rails::Rack::Logger with silencer on sign_in routes
-Rails.application.config.middleware.swap Rails::Rack::Logger, Silencer::Logger, :silence => [%r{^/sign_in}]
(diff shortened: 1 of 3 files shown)


config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,9 @@
 # Be sure to restart your server when you modify this file.
+require 'silencer/logger'
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password, :authenticity_token, :token]
+Rails.application.config.filter_parameters += [:password, :authenticity_token, :token, :email]
+Rails.application.config.action_mailer.logger = nil
```
timwis, line 6: Wouldn't this also silence the logs in development mode? Then we wouldn't be able to get the link from the terminal. Maybe add an `unless ENV['RAILS_ENV'] = 'development'` or something?

timwis: Actually [this looks like a better way](https://stackoverflow.com/questions/49086693/how-do-i-remove-mail-html-content-from-rails-logs):

**config/environments/production.rb**
```ruby
  ...
  config.action_mailer.logger = nil
  ...
end
```

timwis: [What devise recommends](https://github.com/heartcombo/devise#password-reset-tokens-and-rails-logs) may be even better.

rik86: Hi Tim, thanks for the comments I will put it in the production.rb file, I think 'devise' requires more configuration

timwis: I think devise was just saying set Ruby’s log level to WARN to omit the email logs (info). Not sure which approach is better. Probably the first one lol

PR #57, opened 2020-04-02
Add specs for needs controller

jamiebuckley:


spec/controllers/needs_controller_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  describe "GET #index" do
+    it "orders by created_at in descending order by default" do
+      expect(need).to receive(:order).with(:created_at => :desc).and_return(need)
+      get :index
+    end
+
+    it "filters on page number passed in params" do
+      expect(need).to receive(:page).with("5").and_return(need)
+      get :index, params: { page: 5 }
```
rik86, line 24: could you add `expect(response).to be_successful` after the get action, it seems a bit odd we don't verify anything after the call

timwis:
FYI, haven't tested this branch, but when adding #65 I was able to filter and sort by `is_urgent` without the param being in the [accepted params list](https://github.com/wearefuturegov/beacon/blob/master/app/controllers/needs_controller.rb#L9). I think that's because we're using `params.permit()` when we should be using `params.permit!()` but I didn't look very closely for sake of time.

PR #66, opened 2020-04-03
add page size to needs table

rik86:
image

### DIFF
--- a/app/controllers/needs_controller.rb
+++ b/app/controllers/needs_controller.rb
@@ -20,8 +20,8 @@ class NeedsController < ApplicationController
     else
       @needs = @needs.order(created_at: :desc)
     end
-
-    @needs = @needs.page(params[:page])
+    page_size = params[:page_size].to_i > 0 ? params[:page_size] : 10
+    @needs = @needs.page(params[:page]).per(page_size)
   end
 
   def show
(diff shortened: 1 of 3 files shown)


app/controllers/needs_controller.rb
```diff
@@ -20,8 +20,8 @@ def index
     else
       @needs = @needs.order(created_at: :desc)
     end
-
-    @needs = @needs.page(params[:page])
+    page_size = params[:page_size].to_i > 0 ? params[:page_size] : 10
```
timwis, line 23: 1. We'll want to add `:page_size` to `params.permit` above to avoid console errors
2. This sets `page_size` as a string if one is provided, and a number if one isn't
3. This allows page sizes of any number (e.g. 99999999999) rather than restricting it to the 4 options in your helper

app/views/needs/index.html.erb
```diff
@@ -4,9 +4,20 @@
   <h1 class="page-title">Needs <span class="page-title__count">(<%= @needs.total_count %>)</span></h1>
 </div>
 
+     
+
+<%= form_with(url: needs_url, method: :get, local: true) do %>
```
timwis, line 9: I think you may have moved this outside of `<details>` accidentally. It may well break the styling, and the `end` tag (which renders `</form>`) is still inside of `<details>` so would produce invalid markup.

timwis:
UI suggestion but what about moving page size to beneath the table? Not sure it makes sense in the filtering boxes (not to mention it pushing them to two lines) :/

PR #67, opened 2020-04-03
Add export csv link

rik86:
image

### DIFF
--- a/app/controllers/needs_controller.rb
+++ b/app/controllers/needs_controller.rb
@@ -21,28 +21,12 @@ class NeedsController < ApplicationController
     else
       @needs = @needs.order(created_at: :desc)
     end
-    page_size = params[:page_size].to_i > 0 ? params[:page_size] : 10
-    @needs = @needs.page(params[:page]).per(page_size)
+    @needs = @needs.page(params[:page]) unless request.format == 'csv'
 
-  end
-
-  def export
-    @params = params.permit(:user_id, :status, :category, :page, :order_dir, :order, :commit)
-    @users = User.all
-    @needs = Need.includes(:contact, :user)
-    @needs = @needs.filter_by_category(params[:category]) if params[:category].present?
-    @needs = @needs.filter_by_user_id(params[:user_id]) if params[:user_id].present?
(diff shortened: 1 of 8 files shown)


app/controllers/needs_controller.rb
```diff
@@ -20,8 +20,28 @@ def index
...
       @needs = @needs.order(created_at: :desc)
     end
+    page_size = params[:page_size].to_i > 0 ? params[:page_size] : 10
+    @needs = @needs.page(params[:page]).per(page_size)
 
-    @needs = @needs.page(params[:page])
+  end
+
+  def export
+    @params = params.permit(:user_id, :status, :category, :page, :order_dir, :order, :commit)
```
timwis, line 29: This section should include the `:is_urgent` bits from my PR if we want to allow filtering/sorting by that param too.

app/views/needs/index.html.erb
```diff
@@ -4,9 +4,19 @@
...
+     
+
+<%= form_with(url: needs_url, method: :get, local: true) do %>
+
 <details class="filters">
   <summary class="filters__summary">Search and filter</summary>
-  <%= form_with(url: needs_url, method: :get, local: true) do %>
+
+  <!-- Page size -->
+  <div style="width: 100px">
```
timwis, line 15: Is this still necessary if we have CSV export?

config/routes.rb
```diff
@@ -4,6 +4,8 @@
   end
   get '/contacts/:id/needs', to: 'contacts#show_needs'
 
+  get '/needs/export', to: 'needs#export'
```
timwis, line 7: I think this could just be adding `:export` to the array below, no?

app/views/needs/index.html.erb
```diff
@@ -4,9 +4,19 @@
   <h1 class="page-title">Needs <span class="page-title__count">(<%= @needs.total_count %>)</span></h1>
 </div>
 
+     
+
+<%= form_with(url: needs_url, method: :get, local: true) do %>
```
timwis, line 9: Why was this moved up? I worry this would break the styling, no? And wouldn't the close tag need to be moved too?

PR #68, opened 2020-04-03
Fix page size dropdown

rik86:


timwis:
If we merge #67 I think we'd close this without merging

PR #69, opened 2020-04-03
94-add-ci-testing

rubyforbusiness:
I've tried this out extensively on a temporary branch.
It works by putting a file in .github/workflows which is trigged by the action specified at the top, in this case, push to master. The actions are then run in a containerised environment. See the Actions button on the github tabs.

Merging this into master should not prevent any code being merged.
But it will reveal (in Actions) whether any specs are failing.

When we're ready, we can put in place some 'required status checks' on the master branch that call this workflow and reject the merge if it fails - see https://help.github.com/en/github/administering-a-repository/enabling-required-status-checks.


.github/workflows/run_spec_tests.yml
```diff
@@ -0,0 +1,65 @@
+name: Ruby Test Env
+# As opposed to the ci env used before on the Ruby action
+
+on:
+  push:
+    branches: [ master ]
```
timwis, line 6: Let's definitely add the 'on PR' option

.github/workflows/run_spec_tests.yml
```diff
@@ -0,0 +1,65 @@
...
+        bundle install --jobs 4 --retry 3
+    - name: Setup database
+      env:
+        RAILS_ENV: test
+        POSTGRES_USER: test_local_user
+        POSTGRES_PASSWORD: password
+        POSTGRES_DB: i_have_i_need_test
+        POSTGRES_HOST: localhost
+        POSTGRES_PORT: ${{ job.services.postgres.ports[5432] }}
+        SECRET_KEY_BASE: fcd3d03a727ad50a19d2a89555bc9683
```
timwis, line 50: Is this a secret? If so please alert the team and reset it wherever else we use it now that it's been pushed. If not, maybe we should replace it with a string like `test_secret` so it's clear that it's not sensitive?

.github/workflows/run_spec_tests.yml
rubyforbusiness: Done.

.github/workflows/run_spec_tests.yml
rubyforbusiness: This secret was generated by me specifically for the test environment. I can't see how it would be used anywhere else. Anyway I've been refining the script and it turns out you don't need any of the environment vars including the secret. (In an earlier version I was trying to use a separate ci environment but that failed). 
So I've removed it.

.github/workflows/run_spec_tests.yml
rubyforbusiness: Just seen this actually happen on updating this PR: https://github.com/wearefuturegov/beacon/actions/runs/69970929

PR #74, opened 2020-04-06
166 change log level production

rik86:
- ensure only WARN logs are shown in production mode, avoiding printing sensitive information, like user names and magic links, as well as email contents
- adds a formatter for logs so that timestamp and log levels are also shown correctly.

Eg:

image


config/environments/production.rb
```diff
@@ -56,7 +56,11 @@
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :debug
+  Rails.logger = Logger.new(STDOUT)
+  Rails.logger.level = Logger::INFO
```
timwis, line 60: Shouldn't this be `WARN`? I actually don't understand why rails would make you specify the log level twice (once in `Rails.logger.level` and once in `config.log_level`) but it may just be my lack of familiarity with rails.

timwis:
Didn't the previous PR also filter out the `:token` param from the `/sign-in/` page access logs? I don't see that in here. Otherwise looks fine to me but I'm not familiar with changing log settings in rails.

config/environments/production.rb
rik86: Hi @timwis  regarding your comment, you are right, I did set it to "WARN" at line 63. so we would not print anything below WARN in production.

Line 60 defines the default level for rails messages, I was referencing this guide: https://stackify.com/rails-logger-and-rails-logging-best-practices/

rik86:
> Didn't the previous PR also filter out the `:token` param from the `/sign-in/` page access logs? I don't see that in here. Otherwise looks fine to me but I'm not familiar with changing log settings in rails.

That would only work for query params, like `/sign-in/?token=whatever` but does not work with path urls like `/sign-in/whatever` that's why I was trying to use the `silencer` gem to hide routes logs. But that does not really matter now as we settled to hide anything not important on PROD and display everything as normal for DEV mode....right?

rik86:
[Fix Logs with magic link](https://trello.com/c/RtcIUvMg/166-fix-logs-with-magic-link)

timwis:
@rubyforbusiness the goal of this PR is to solve two problems:

1. Outbound "finish logging in" emails are being logged in production server logs, which include the magic links
2. When people click those magic links, they show up in the production server _access_ logs (as they're part of the URL)

It sounds like @rik86 may have fixed the second one in a different request somewhere?

I defer to your recommendation (or whatever rails' default is) on production log levels and the best way to solve them.

rubyforbusiness:
> @rubyforbusiness the goal of this PR is to solve two problems:
> 
> 1. Outbound "finish logging in" emails are being logged in production server logs, which include the magic links
> 2. When people click those magic links, they show up in the production server _access_ logs (as they're part of the URL)
> 
> It sounds like @rik86 may have fixed the second one in a different request somewhere?
> 
> I defer to your recommendation (or whatever rails' default is) on production log levels and the best way to solve them.

Thanks for the clarity.
So the only quick way to stop the magic links being seen in both the GET requests , and the controller params, is to set the level to WARN, warn. So that's what I've done. It solves the immediate problem. But it does give very little logging in production. If that's an issue we can create a new ticket to address it, then install some kind of logging filter (e.g. maybe https://github.com/cookpad/blouson).

PR #75, opened 2020-04-06
Support review apps

timwis:
This will allow us to run `db:seed` in production without loading factorbot+faker when the server runs, and that command will seed the users table with any emails specified in the env var. And the hostnames in emails will work on review apps.

Just need to add a command to the heroku config so that review apps run `db:seed` when they're created, but that can be a separate PR.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -44,9 +44,9 @@ end
 # make an initial user for sake of the readme
 seed_user_emails = ENV['SEED_USER_EMAILS'] || 'admin@example.com'
 seed_user_emails.split(',').each do |email|
-  User.create(
-    email: email.strip,
-    admin: true,
-    invited: DateTime.now
-  )
+  User.find_or_initialize_by(email: email.strip)
+    .update!(
+      admin: true,
+      invited: DateTime.now
+    )
 end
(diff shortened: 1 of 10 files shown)


timwis:
@rubyforbusiness any reason the CI uses `bundle exec rails spec` instead of `bundle exec rspec`? As you pointed out, tests pass with the latter, and rspec-rails docs suggest using the latter. I still think it's weird that either would fail, but wanted to check where the former command came from.

timwis:
It's working!! Muhuhahaha :D I created a review app in heroku off of this branch, and it had seed data and my user account in there when it finished creating.

db/seeds.rb
```diff
@@ -39,10 +42,11 @@
...
-  email: "admin@example.com",
-  first_name: "Example",
-  last_name: "User",
-  admin: true,
-  invited: "2020-03-25 00:00:00"
-)
\ No newline at end of file
+seed_user_emails = ENV['SEED_USER_EMAILS'] || 'admin@example.com'
+seed_user_emails.split(',').each do |email|
+  User.create(
```
timwis, line 47: Actually I’d like to use User.find_or_initialize_by(email: email.strip).update!(admin: true, invited:DateTime.now) so it won’t error if you run seed a second time.

spec/mailers/user_sign_in_mailer_spec.rb
```diff
@@ -1,6 +1,11 @@
 require "rails_helper"
 
 RSpec.describe UserSignInMailer, type: :mailer do
+
+  before do
```
jamiebuckley, line 5: To explain this for posterity
See issue 1644 of https://github.com/rspec/rspec-rails/
```
routes.draw
```
combined with parallel tests causes routes to drop

PR #78, opened 2020-04-06
Feature/notify

jhackett1:
Replace sendgrid with notify for the invitation and magic link emails

Notify API keys have been added to review, staging, qa and prod camden environments ready.

The notify template IDs are config from `councils.yml`.

**The rspec tests are broken and i'm not sure how to fix them.** . They're fixed now.


Gemfile
```diff
@@ -16,19 +16,17 @@ gem 'rails', '~> 6.0.2', '>= 6.0.2.2'
 gem 'sass-rails', '>= 6'
 gem 'turbolinks', '~> 5'
 gem 'webpacker', '~> 4.0'
-gem 'sendgrid_actionmailer_adapter'
+# gem 'sendgrid_actionmailer_adapter'
 gem 'strong_migrations'
 gem 'csv'
-
-group :development, :test do
-  gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
```
rik86future, line 24: I use `byebug` in development too, it'd be great if it could stay there

jhackett1: me too! it's still in there, i just moved it into the other development/test group further down the file

timwis:
We're still discussing the best way to proceed on the email provider issue, so hold off merging this until we decide (hopefully within the hour).

rubyforbusiness:
> We're still discussing the best way to proceed on the email provider issue, so hold off merging this until we decide (hopefully within the hour).

OK. Looking at the types of changes involved (a lot of config) I don't think it would be too hard to merge the rubocop changes back into this - but we'll avoid that unless we have to.

spec/mailers/user_sign_in_mailer_spec.rb
```diff
@@ -2,19 +2,30 @@
...
-    let(:mail) { described_class.send_invite_email(user).deliver }
+    let(:mail) { described_class.send_invite_email(user, 'http://localhost:3000/') }
+    let(:personalisation) { { beacon_url: 'http://localhost:3000/' } }
+
+    it 'sets the recipient' do
+      expect(mail.to).to eq(['test@emailaddress.com'])
+    end
+    
+    it 'sets the subject' do
+      expect(mail.body).to match("This is a GOV.UK Notify email with template #{template}")
```
timwis, line 16: Where does this body come from?

spec/mailers/user_sign_in_mailer_spec.rb
```diff
@@ -2,19 +2,30 @@
 
 RSpec.describe UserSignInMailer, type: :mailer do
   describe 'send invite email' do
+
+    let(:template) { '31d2b530-8db4-44ce-9bfa-34b01519335f' }
```
timwis, line 6: Ideally we'd use a dummy value like 'example-template' or pull the default value from councils.yml, because if we change the value in councils.yml later, this test will fail. But not a big deal.

config/environments/test.rb
```diff
@@ -6,6 +6,8 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  config.action_mailer.default_url_options = { host: ENV['HOSTNAME'] || 'localhost:3000' }
```
timwis, line 9: I did this in #75 so we may have a merge conflict depending on which gets merged first. Or this line may be in there twice if we put them at different parts of the file.

config/councils.yml
```diff
@@ -22,6 +26,8 @@ default: &default
     logo_path: "logo-camden.svg"
     privacy_link: "https://www.camden.gov.uk/data-protection-privacy-and-cookies"
     default_from_address: "help@camden.beacon.support"
+    invite_template: '31d2b530-8db4-44ce-9bfa-34b01519335f'
```
timwis, line 29: Are we just using Camden's template for everything at the moment? If so, we should probably just use the Beacon logo instead of the Council's logo so that we can demo something neutral.

Gemfile
```diff
@@ -16,19 +16,17 @@ gem 'rails', '~> 6.0.2', '>= 6.0.2.2'
 gem 'sass-rails', '>= 6'
 gem 'turbolinks', '~> 5'
 gem 'webpacker', '~> 4.0'
-gem 'sendgrid_actionmailer_adapter'
+# gem 'sendgrid_actionmailer_adapter'
```
timwis, line 19: There's a few commented out lines in here. It would be great if those lines could be deleted and we could just use version history to get them back later if need be, but it's not a big deal if we're pressed for time.

config/councils.yml
jhackett1: we can't use beacon's logo because the logo is tied to the notify account we're using, which is in turn tied to the email address of whoever creates the account (camden emma in this case)

PR #83, opened 2020-04-07
fix broken logs

rik86future:


timwis:
I think all 3 of the lines that were added in #74 need to be removed from development.rb, otherwise it looks like you get the silence error that causes styles not to render.

PR #84, opened 2020-04-07
Spike: Add jsonb column to needs and use STI

timwis:
This is just exploratory, and one of several approaches we'll discuss tomorrow, not necessarily the best one. I was just curious what it could look like, and I'm sure it could be approved if we even decide to go this direction.

This PR adds a jsonb column to the needs table and uses Single Table Inheritance  to handle the need types having their own fields. This does mean hardcoding the need types rather than letting them be dynamic, although we were already doing that to an extent, just as a list rather than as models.

Note that if you want to run this, you'll probably need to comment out the three `Logger` lines in config/environment/development.rb until #83 is resolved.

Screenshot 2020-04-07 at 20 31 54
Screenshot 2020-04-07 at 20 32 03
Screenshot 2020-04-07 at 20 32 15
Screenshot 2020-04-07 at 20 32 23

### DIFF
--- a/app/models/needs/need.rb
+++ b/app/models/needs/need.rb
@@ -4,6 +4,7 @@ require 'csv'
 
 class Need < ApplicationRecord
   include Filterable
+  self.inheritance_column = 'category'
 
   belongs_to :contact, counter_cache: true
   belongs_to :user, optional: true
--- a/config/application.rb
+++ b/config/application.rb
@@ -15,6 +15,8 @@ module IHaveINeed
     # <%= Rails.configuration.councils[ENV['COUNCIL']]['support_email'] %>
     config.councils = config_for(:councils)
 
+    config.autoload_paths += %W(#{Rails.root}/app/models/needs)
+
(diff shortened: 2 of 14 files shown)


app/models/needs/need.rb
```diff
@@ -4,6 +4,7 @@
 
 class Need < ApplicationRecord
   include Filterable
+  self.inheritance_column = 'category'
```
timwis, line 7: If the column were called `type`, we wouldn't even need this line, as rails assumes STI by default with that field name, apparently.

config/application.rb
```diff
@@ -15,6 +15,8 @@ class Application < Rails::Application
     # <%= Rails.configuration.councils[ENV['COUNCIL']]['support_email'] %>
     config.councils = config_for(:councils)
 
+    config.autoload_paths += %W(#{Rails.root}/app/models/needs)
```
timwis, line 18: This is just so we can put the need models in their own folder for housekeeping, but is optional.

timwis:
Shuddup, rubocop

PR #89, opened 2020-04-08
Spike on Call log [JUST FOR REFERENCE] don't merge

rik86future:
image

image


rik86future:
[Spike Log call functionality](https://trello.com/c/xf0ihJBy/224-spike-log-call-functionality)

PR #92, opened 2020-04-09
Feature/secured admin area

jhackett1:
- secure the user manager to non-admins
- hide the nav link to the user manager for non-admins

- fix a bug with last_logged_in not displaying

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -46,6 +46,7 @@ class UsersController < ApplicationController
 
   def require_admin
     return if current_user.admin
+
     redirect_to root_path, flash: { error: 'Only administrators can manage users' }
   end
 end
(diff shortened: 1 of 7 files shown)


app/controllers/users_controller.rb
```diff
@@ -43,4 +43,10 @@ def set_user
   def user_params
     params.require(:user).permit(:email, :admin)
   end
+
+  def require_admin
+    return if current_user.admin
```
timwis, line 48: Wouldn't this return a falsy value if the user is an admin? I would have expected `return true if current_user.admin` but maybe I'm just not used to ruby

timwis:
LGTM apart from that one question I left

app/controllers/users_controller.rb
eddieleeper-nb: it always returns `nil` but the redirect isn't executed because the function returns

timwis: ah right, makes sense. it's "require admin" not "is admin"

PR #93, opened 2020-04-09
Feature/229 upgrade the add needs page

jamiebuckley:


app/assets/images/beacon-white.svg
```diff
@@ -0,0 +1,36 @@
+<svg width="136" height="45" viewBox="0 0 136 45" fill="none" xmlns="http://www.w3.org/2000/svg">
```
jamiebuckley, line 1: This is just the white beacon svg from the designs

app/controllers/contact_profiles_controller.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
```
jamiebuckley, line 1: Extracted this from the mixed contacts/needs controller, as it is getting confusing.  Makes it easier to rewrite again when we do the full v2 designs

app/views/contacts/_form.html.erb
```diff
@@ -1,161 +1,6 @@
...
-      </div>
-      <div class="radio">
-        <%= form.radio_button :is_vulnerable, false, :id => "is_vulnerable_false" %>
-        <label for="is_vulnerable_false" class="radio__label">No</label>
-      </div>
-
-    </fieldset>
-
-
+    <%= render partial: 'form_fields', locals: { contact: @contact, form: form } %>
```
jamiebuckley, line 3: This is copied verbatim, no changes.  The purpose is to be able to embed the fields from the contact form inside the combined needs/profile edit form

timwis:
Looks like there's a CSS issue on firefox
Screenshot 2020-04-09 at 12 09 32

jamiebuckley:
@timwis yep I suspected that.  There are quite a lot of css issues with the navbar, debating leaving it as is, as chances are @jhackett1 is going to rewrite it

timwis:
Okay. Maybe set `display: none` on it for now so it's at least functional, and Jaye can unhide it when they can fix it..

Also looks like the radios are not on the same line, though not a huge deal:

Screenshot 2020-04-09 at 12 11 03

More significantly, I thought we were going to put the new food related questions as part of the food need, and the new prescription question(s) as part of the prescription need? (even though under the hood they're stored on the contact model)

jamiebuckley:
The issue is that it embeds the edit profile form, and those questions are now included in that form.  I can potentially not use the contact edit form, and add a new form for this page

timwis:
@jamiebuckley ah, I see. Another option, considering agents would only really be updating those new extra questions _while in the process of adding a need_, would be to remove them from the shared edit profile page and only have them on the add need page 🤔

jamiebuckley:
@timwis That does sound better, I agree.  Will change it to that

timwis:
@jamiebuckley Yet another option is for the shared "edit contact form" to only include true contacty stuff, and have partials for the extra questions (food questions partial, prescription questions partial). On the "edit contact" page, include the edit contact form partial, and below that the extra questions partials. On the new "call agent edit form", include the "edit contact" partial at the top, then the extra question partials in the context of the relevant need.

app/controllers/contact_profiles_controller.rb
timwis: So this is the new "update contact and add needs" controller that contact centre agents will use, right?

config/routes.rb
```diff
@@ -3,10 +3,12 @@
     resources :needs, only: [:new, :create]
   end
   get '/contacts/:id/needs', to: 'contacts#show_needs'
+  resources :contact_profiles, only: [:edit, :update]
```
timwis, line 6: Hm, not sure about the name for this... Probably makes more sense as a nested route under the `/contacts` resource (although I know it maps to a different controller), like `/contacts/123/triage` or something. Maybe we call it the triage controller while we're at it?

jamiebuckley: Yep that's right, it didn't seem to make sense for it to be mixed with the needs_controller functionality any more.  I struggled for what to name it though.  Triage sounds good though, will change it to that.

timwis:
I'm hopping on a call but that was my only piece of feedback so if that's resolved you can consider the PR to have at least 1 approval!

PR #98, opened 2020-04-09
Add note types for calls (don't merge before UI overhaul)

rubyforbusiness:
This adds configurable categories to notes per need category and uses radio buttons to input them.


rubyforbusiness:
Sorry didn't have time to write any cukes but I added a couple of rspecs.
The 'hash' implementation isn't great. I tried converting it to a YAML config, but ran out of time (it switches the string keys to symbols, which doesn't work with the code in needs/show). Good luck :rocket:

app/helpers/needs_helper.rb
```diff
@@ -18,8 +18,64 @@ def need_categories
...
+      &.find { |nc| nc[:id] == note_category_id }
+      &.fetch(:display)
+  end
+
+  def needs_note_categories
+    {
+      'Phone triage' => {
+        order: 1,
+        stored_as: 'phone triage',
+        note_categories: [
```
timwis, line 37: I haven't heard a need for note types to be based on the type of need 🤔 I imagine we may need this eventually, but in the meantime it might be over optimised, and we could simplify things a lot by simply adding an optional category for any note.

app/views/needs/show.html.erb
```diff
@@ -96,6 +96,20 @@
   </h2>
 
   <%= form_with(model: [@need, @need.notes.build], local: true) do |form| %>
+    <div class="field">
+      <% nnc = needs_note_categories.find {|e| e[1][:stored_as] == @need.category}  %>
```
timwis, line 100: Let's have a look at the figma designs and see how they handle rendering this so @jhackett1 doesn't have to fix it up if we can handle it. Also note that choosing a category should be optional (hopefully that's reflected in the designs), so you can log a note that isn't about a phone call.

app/helpers/needs_helper.rb
```diff
@@ -18,8 +18,64 @@ def need_categories
...
+          # prefix 'phone_' makes reporting easier
+          { id: 'phone_success', display: 'Successful Phone call' },
+          { id: 'phone_failure', display: 'Failed Phone call' }
```
timwis, line 40: There are a couple other categories, such as left message and interrupted. Siobhan should have the exact list and language.

app/helpers/needs_helper.rb
```diff
@@ -22,6 +22,21 @@ def needs
...
 
+  def note_category_display(note_category_id)
+    note_category = needs_note_categories.find { |category| category[:id] == note_category_id }
+    note_category.present? ? note_category[:display] : 'Note'
+  end
+
+  def needs_note_categories
+    [
+      # prefix 'phone_' makes reporting easier
+      { id: 'note', display: 'Note' },
```
timwis, line 33: Does the fact that other categories can also have a note/description make this id less clear? I wonder if an empty/null category makes more sense? Not a big deal either way.

app/helpers/needs_helper.rb
```diff
@@ -22,6 +22,21 @@ def needs
      'Staying Social', 'Prescription pickups', 'Book drops and entertainment', 'Dog walking']
   end
 
+  def note_category_display(note_category_id)
+    note_category = needs_note_categories.find { |category| category[:id] == note_category_id }
```
timwis, line 26: Doesn't rails have built in enums that we can use for this instead? e.g. `enum category: { 'Successful call': 'phone_success', 'Left message': 'phone_message', ... }` (on the model) Or is this something different?

app/controllers/notes_controller.rb
```diff
@@ -3,13 +3,13 @@
...
-    @note = @need.notes.create(note_params.merge(user: current_user))
+    @note = @need.notes.create!(note_params.merge(user: current_user))
     redirect_to need_path(@need)
   end
 
   private
 
   def note_params
-    params.require(:note).permit(:body)
+    params.require(:note).permit(:body, :category)
```
timwis, line 13: Per slack convo, it probably makes sense to require _either_ of these fields, rather than both. So you can submit with category or with body or with category and body but not without either. But I wouldn't call it a blocker necessarily; requiring both is better than not requiring either.

app/helpers/needs_helper.rb
eddieleeper-nb: I thought enums in the model mapped to integers in the database, but I've just tried what you suggested and it seems to work - nice!

app/controllers/notes_controller.rb
eddieleeper-nb: Tim, only :note is required, :body and :category are permitted if present.

timwis: Okay 🤔 I don't get how `note` is a field on a `Note`, on the same level as `body` and `category`, but maybe I'm missing something. I'm sure it's fine.

app/models/note.rb
```diff
@@ -4,5 +4,10 @@ class Note < ApplicationRecord
   belongs_to :need
   belongs_to :user
 
+  enum category: { 'Note': 'note',
+                   'Successful Call': 'phone_success',
+                   'Left Message': 'phone_message',
+                   'Failed Call': 'phone_failure' }
```
timwis, line 10: Heads up that the database will contain needs of category `phone_import`. We'll need to render those with a friendly name like "Imported call log", but you shouldn't be able to add a _new_ note of type `phone_import`. We can leave that to a trello card [like this one](https://trello.com/c/LUGa0GGa/281-improve-rendering-of-imported-note-contents) so it doesn't block merging this PR though.

app/views/needs/show.html.erb
```diff
@@ -10,12 +10,12 @@
 
 <ol class="breadcrumb">
   <li><%= link_to "People in need", contacts_path %></li>
-  <li><%= link_to @need.contact.name , @need.contact %></li>
+  <li><%= link_to @need.contact.name, @need.contact %></li>
   <li>Need</li>
```
timwis, line 14: Heads up that this file will probably have a lot of conflict's with Jaye's overhaul branch. Probably better to merge that branch before this one and resolve the conflict here.

app/controllers/notes_controller.rb
```diff
@@ -3,13 +3,16 @@
 class NotesController < ApplicationController
   def create
     @need = Need.find(params[:need_id])
-    @note = @need.notes.create(note_params.merge(user: current_user))
+    params = note_params
+    params.merge!(body: 'No details captured') if note_params[:body].blank?
```
timwis, line 7: This is probably fine for the time being, if you create a trello card in the backlog about sorting the rules out for this. I think ideally we'd require either a category (assuming non phone calls have an empty category; otherwise we'd require a non-"note" category) _or_ body. But it's fine for now.

app/models/note.rb
```diff
@@ -4,5 +4,10 @@ class Note < ApplicationRecord
   belongs_to :need
   belongs_to :user
 
+  enum category: { 'Note': 'note',
```
timwis, line 7: Is there any reason to store `'note'` in the database, instead of leaving it blank/Nil?

app/controllers/notes_controller.rb
eddieleeper-nb: Trello card created.

app/controllers/notes_controller.rb
eddieleeper-nb: it's fine :-)

app/models/note.rb
eddieleeper-nb: Just seemed like a more concrete abstraction. I'll update to be nil

app/models/note.rb
eddieleeper-nb: ok - noted.

app/views/needs/show.html.erb
eddieleeper-nb: agreed - done

spec/models/need_spec.rb
```diff
@@ -17,11 +17,11 @@
     let!(:uncompleted_need) { create :need }
     let!(:completed_need) { create :need, :completed }
 
-    it '.completed' do
-      expect(described_class.uncompleted).to eq [uncompleted_need]
+    it '.uncompleted' do
```
eddieleeper-nb, line 20: I'm not sure how this is working in the pipeline or on others peoples machines - this broke for me so applying a fix.  If this is not correct please let me know and I'll revert.

app/views/needs/_notes.html.erb
```diff
@@ -2,6 +2,16 @@
   
   <%= form_with(model: [@need, @need.notes.build], local: true, class: "notes__form") do |form| %>
     <h3 class="notes__title">Updates</h3>
+    <div class="field">
+      <%= form.label :category, 'Choose a category', class: "field__label" %>
+      <% Note.categories.each.with_index do |nc, index| %>
```
timwis, line 7: Rails has a form helper [`collection_radio_buttons`](https://apidock.com/rails/v4.0.2/ActionView/Helpers/FormOptionsHelper/collection_radio_buttons) that could simplify this and eliminate the need for the helpers above. But I couldn't get it to set the default option to general/note.

```ruby
      <%= form.collection_radio_buttons(:category, Note.categories, :last, :first) do |b| %>
        <div class="radio">
          <%= b.radio_button %>
          <%= b.label(class: 'radio__label') %>
        </div>
      <% end %>
```

No big deal either way - your code works and LGTM

PR #100, opened 2020-04-10
fix failing rspec tests

eddieleeper-nb:


eddieleeper-nb:
false alarm - suspect spring application pre-loader

PR #103, opened 2020-04-10
Import Camden data

timwis:
~~I revised the contacts import logic in [my python script](https://github.com/timwis/beacon-data-importer) and then realised when writing the call log importer that they'd need to reference contacts (and notes would need to reference needs), so it's much easier to use activerecord after all, despite the fact that it's less efficient. @eddieleeper-nb called it.~~

~~Trouble is I haven't been able to test this yet, primarily because our seed data doesn't have all the fields we need (especially the json data). The synthetic data doesn't have referential integrity between Shielded IDs. And frankly my brain is about fried.~~

~~So this is ruby code that I haven't actually tested. Is there a way to output the active record model in the terminal before it persists to the database? I hope this is close to being right. I could use a second pair of eyes in the morning if anyone's around.~~

~~You should theoretically be able to run it via:~~

```bash
rake camden_data:prepare_contacts[path/to/healthintent.csv,path/to/gds.csv]
rake camden_data:prepare_calls[path/to/calls.csv]
```

Update 11/4: I've removed the rake task and am doing the import via python cleanup scripts and SQL scripts. Now this PR is simply the migrations needed to support the new data. I'm deploying it to their prod environment now so I can import the data. So hopefully we can merge this once reviewed.


timwis:
I've added a few more commits to (1) simplify and update the seeds to reflect our current data model, and (2) add a camden-specific seed (run it via `COUNCIL=camden rails db:seed`) that resembles what we expect them to have on day 1.

db/seeds.rb
```diff
@@ -14,23 +14,25 @@
...
-    FactoryBot.create_list(:user, 5, organisation: organisation).each do |user|
-      contact_list = FactoryBot.create :contact_list
-
-      FactoryBot.create :contact_list_user, contact_list: contact_list, user: user
-
-      contacts = FactoryBot.create_list :contact, 50, contact_list: contact_list
-
-      need_categories = ['phone triage', 'groceries and cooked meals', 'physical and mental wellbeing', 'financial support',
-                        'staying social', 'prescription pickups', 'book drops and entertainment', 'dog walking', 'other']
+  if ENV['COUNCIL'] == 'camden'
```
eddieleeper-nb, line 17: Might be worth adding a bit to the README.md about this ENV var

spec/factories/notes.rb
```diff
@@ -1,9 +1,17 @@
 # frozen_string_literal: true
 
+note_categories = [
+  'note',
```
eddieleeper-nb, line 4: this is now `general` after our slack discussion this morning.

eddieleeper-nb: Ah but I think this is where the merge conflict is so will get resolved when the merge conflict gets resolved.

db/seeds.rb
timwis: We have got it in the list of env vars in the readme - did you mean somewhere else?

eddieleeper-nb: I meant adding a line about running the seeds with that env var, but it's probably fine without it.  `COUNCIL=camden rails db:seed` as per your comment on the PR

PR #104, opened 2020-04-11
Fix 'Edit' link on personal details to the go to the triage page

go8soft:
Fix 'Edit' link on personal details to the go to the triage page

### DIFF
--- a/app/views/shared/_profile-accordion.html.erb
+++ b/app/views/shared/_profile-accordion.html.erb
@@ -23,7 +23,7 @@
                 <%= @contact.mobile %>
             </dd>
         </dl>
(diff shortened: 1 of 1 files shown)


jhackett1:
given that we already have the edit person fields on a separate form by virtue of using them on the triage form, why do we need to deliberately turn it off? why not just leave it be?

i suspect trying to do everything through the triage form will be confusing to the user since it straddles two different concepts and is somewhere between a create and an edit form

timwis:
Does the edit profile page work? If so, I agree, leave it be. It was more about prioritising the triage page, and on the old design it was confusing to have two different buttons. This design is less confusing. If the edit profile page works, I agree let's close this PR.

jhackett1:
it does indeed work. will close.

PR #107, opened 2020-04-11
Fix: email link on sign in page

adragolov:
Background:
https://trello.com/c/QjcqfFL6/271-fix-email-link-on-sign-in-page

Solution:

- The value for the support email is provided by the Rails configuration and exposed to the Application controller
- Layout for the page is fixed, value is rendered on the server


adragolov:
[Fix email link on sign in page](https://trello.com/c/QjcqfFL6/271-fix-email-link-on-sign-in-page)

PR #108, opened 2020-04-11
Fix needs filters and assigned to me link

go8soft:
- Fix needs filters
- Fix Assigned to me link. Now we can use filters with assign to me logic.
- Add redirect in routes. Because of that we can use link_to_unless_current correctly.

Task cards:

https://trello.com/c/wrKKOoWj/269-keep-navigation-indicators-selected-when-paging-through-or-sorting-need-index-view

https://trello.com/c/mIIZ78Vm/278-make-assigned-to-me-nav-link-work-again

Before merge need to talk about pagination problem and generally do we need navs or filters.

### DIFF
--- a/app/controllers/needs_controller.rb
+++ b/app/controllers/needs_controller.rb
@@ -8,10 +8,10 @@ class NeedsController < ApplicationController
   helper_method :get_param
 
   def index
-    @params = params.permit(:user_id, :status, :category, :page, :order_dir, :order, :commit, :is_urgent)
-    @users = User.all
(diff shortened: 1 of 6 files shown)


timwis:
Looks like a good implementation! Haven't tested it locally, but I'm wondering if perhaps the need to support the word "mine" requires too much complexity? We could just go back to doing `?assigned_to=12` (current_user.id) and this would be pretty simple, wouldn't it? If we want something prettier, we could use the email address.

go8soft:
I just had a talk with @timwis . I need to rework the PR.

PR #109, opened 2020-04-13
Fix filtering of needs

go8soft:
Fix filtering of needs:

- Re-enable/fx `assigned to` filter
- Update nav bar's 'assigned to me' link to use `current_user`
- Fix category filter
- Nav bar link highlights should indicate current page

Trello: https://trello.com/c/k9ABWrSy/293-fix-filtering-of-needs

### DIFF
--- a/app/views/needs/_filters.html.erb
+++ b/app/views/needs/_filters.html.erb
@@ -9,16 +9,16 @@
     <% end %>
         <%= form_with(url: root_path, method: :get, local: true, class: "filters__form") do %>
           <label for="user_id" class="visually-hidden">By assigned user</label>
-          <%= select_tag("user_id", content_tag(:option,'Unassigned',:value => "Unassigned", :selected => params["user_id"] == "Unassigned") + options_from_collection_for_select(@users, :id, :name_or_email, :selected => params['user_id']), prompt: 'Assigned to', class: "filters__select") %>
+          <%= select_tag("user_id", content_tag(:option,'Unassigned',:value => "Unassigned", :selected => params["user_id"] == "Unassigned") + options_from_collection_for_select(@users, :id, :name_or_email, :selected => params['user_id']), prompt: 'All users', class: "filters__select") %>
 
           <label for="category" class="visually-hidden">By category</label>
-          <%= select_tag("category", content_tag(:option,'Other',:value => "Other", :selected => params["category"] == "Other") + options_for_select(needs_labels_list, :selected => get_param_capitalized('category')), prompt: 'Category', class: "filters__select") %>
+          <%= select_tag("category", content_tag(:option,'Other',:value => "Other", :selected => params["category"] == "Other") + options_for_select(needs_labels_list, :selected => get_param_capitalized('category')), prompt: 'All categories', class: "filters__select") %>
         
           <label for="status" class="visually-hidden">By status</label>
-          <%= select_tag("status", options_for_select(need_statuses, :selected => get_param_capitalized('status')), prompt: 'Status', class: "filters__select") %>
+          <%= select_tag("status", options_for_select(need_statuses, :selected => get_param_capitalized('status')), prompt: 'All statuses', class: "filters__select") %>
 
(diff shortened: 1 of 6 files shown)


app/views/needs/_filters.html.erb
```diff
@@ -7,19 +7,22 @@
...
+          <%= select_tag("category", content_tag(:option,'Other',:value => "Other", :selected => params["category"] == "Other") + options_for_select(needs_labels_list, :selected => get_param_capitalized('category')), prompt: 'All', class: "filters__select") %>
         
           <label for="status" class="visually-hidden">By status</label>
           <%= select_tag("status", options_for_select(need_statuses, :selected => get_param_capitalized('status')), prompt: 'All', class: "filters__select") %>
 
           <label for="is_urgent" class="visually-hidden">By priority</label>
           <%= select_tag("is_urgent", options_for_select(need_urgencies, :selected => get_param_capitalized('is_urgent')), prompt: 'All', class: "filters__select") %>
 
+          <% if params[:user_id] %>
+            <%= f.hidden_field :user_id, value: params[:user_id] %>
```
timwis, line 24: Please re-enable the "By assigned user" dropdown. With that enabled, we can remove the hidden field.

app/views/needs/_filters.html.erb
```diff
@@ -7,19 +7,22 @@
         <button class="button button--dark filters__controls" aria-expanded="false">Filters</button>
         <div class="filters__content" hidden>
     <% end %>
-        <%= form_with(url: needs_url, method: :get, local: true, class: "filters__form") do %>
+        <%= form_with(url: root_path, method: :get, local: true, class: "filters__form") do |f| %>
```
timwis, line 10: Hm, why change `needs_url` to `root_path`?

app/views/shared/_side-nav.html.erb
```diff
@@ -8,18 +8,26 @@
     <ul class="main-menu">
         <li class="main-menu__item">
             <%= link_to_unless_current "Needs", root_path, class: "main-menu__link" do %>
```
timwis, line 10: Is this line necessary now that we have a `link_to` in the block?

app/views/shared/_side-nav.html.erb
```diff
@@ -8,18 +8,26 @@
...
                 <span class="main-menu__current">Needs</span>
+              <% end %>
             <% end %>
 
             <ul class="main-menu__submenu">
                 <li class="main-menu__submenu-item">
-                    <%= link_to_unless_current "Assigned to me", root_path(:mine => true), class: "main-menu__submenu-link" do %>
-                        <span class="main-menu__submenu-current">Assigned to me</span> 
-                    <% end %>
+                  <% if current_page?(root_path) && params[:user_id] %>
```
timwis, line 20: Wouldn't `link_to_unless_current` work with `?user_id` set? Seems like we're re-implementing its logic here, no?

app/views/needs/_filters.html.erb
go8soft: ok

app/views/needs/_filters.html.erb
go8soft: Because of `link_to_unless_current` we need to use the same as in helper.

app/views/shared/_side-nav.html.erb
go8soft: Yes, `link_to_unless_current` can work with `user_id`. But when we have more params for example `?user_id=<va>&status=<val>` we will lose `link_to_unless_current` behaviour.

app/views/shared/_side-nav.html.erb
go8soft: Yes, It is necessary.  When we are 'in' submenu option we want to see `Needs` selected and also to be able to press the link 'Needs'.

go8soft:
@timwis I did the changes. Also change the labels from https://trello.com/c/ZhfTMmIR/289-consider-adding-labels-to-filter-dropdowns

PR #110, opened 2020-04-13
Fix needs CSV export

go8soft:


go8soft:
Task: https://trello.com/c/6VMBGUJk/273-export-csv-is-broken

PR #112, opened 2020-04-13
Feature/233 support sorting needs by date of most recent call

jamiebuckley:


jamiebuckley:
[Support sorting needs by date of most recent call](https://trello.com/c/ApgFWmov/233-support-sorting-needs-by-date-of-most-recent-call)

app/models/need.rb
```diff
@@ -87,18 +91,32 @@ def status=(state)
...
 
   def self.base_query
-    Need.includes(:contact, :user)
+    sql = "LEFT JOIN (select c.id, max(nt.created_at) from contacts c
+          left join needs n on n.contact_id = c.id
+          left join notes nt on nt.need_id = n.id where nt.category like 'phone_%'
+          group by c.id) as contact_aggregation
+          on contact_aggregation.id = contacts.id"
+
+    Need.joins(:contact, sql).select('needs.*', 'contact_aggregation.max as last_phoned_date')
```
timwis, line 100: I assume this was to get around the n+1 query problem that would have presumably been caused by using ActiveRecord associations?

Also, did you consider whether this makes more sense as an attribute on the contact model?

jamiebuckley: I did, but due to the need to make it work with the needs list, I left it in the Need model for the moment.  What I'd like to be able to do is define the Contact query and field on the Contact model, and then be able to just join the Contact model in.

The bit that makes it complex at the moment is that the contact is already being loaded via a join (as opposed to the root of the query), so it would be more complex to implement than the query above

timwis: Ah, okay. Didn’t get a chance to dive in but don’t let me hold anything up if Eddie gets eyes on it first. I’ve not solved this problem in rails before anyway.

app/models/need.rb
```diff
@@ -87,18 +91,32 @@ def status=(state)
   end
 
   def self.base_query
-    Need.includes(:contact, :user)
+    sql = "LEFT JOIN (select c.id, max(nt.created_at) from contacts c
```
eddieleeper-nb, line 94: Would be good to have test coverage for this.  Not to stop this going through, but we need to make sure we have some rspec or cucumber tests to cover this.

PR #114, opened 2020-04-13
Add print styles to show all profile details on need page when printing

timwis:
This expands all the profile areas when printing, and makes the accordion titles more visible on a white background. I'm sure we could improve the print styles significantly, but for now I wanted to make sure at least those key fields are displayed (assuming a need may be printed out and handed to a driver).

Screenshot 2020-04-13 at 18 29 43
Screenshot 2020-04-13 at 18 29 53


timwis:
[Printing a need should have all accordions expanded (and probably navbar hidden)](https://trello.com/c/na24Ucai/257-printing-a-need-should-have-all-accordions-expanded-and-probably-navbar-hidden)

PR #115, opened 2020-04-13
Disallow search engine indexing via robots.txt

timwis:


timwis:
[Consider adding a robots.txt that disallows all indexing](https://trello.com/c/CTPBEmn5/287-consider-adding-a-robotstxt-that-disallows-all-indexing)

PR #120, opened 2020-04-14
Tests for Feature/233 and allow running tests in firefox & chrome

jamiebuckley:
Capybara uses https://github.com/rack/rack-test by default, this allows running in selenium + firefox


features/step_definitions/need_steps.rb
```diff
@@ -31,5 +31,6 @@ def yes_selected_id_from(need)
                 prescription_pickups: 'needs_active_5_true',
                 book_drops_and_entertainment: 'needs_active_6_true',
                 dog_walking: 'needs_active_7_true' }
-  yes_needs.fetch(need.downcase.gsub(' ', '_').to_sym)
+  radio_id = yes_needs.fetch(need.downcase.gsub(' ', '_').to_sym)
+  page.find("label[for=#{radio_id}]").click
```
jamiebuckley, line 35: In a real browser, the labels are considered to obscure the input, hence 'choose' does not work.  Finding and clicking the label solves those

PR #121, opened 2020-04-14
301 bugfix for user display in needs list

eddieleeper-nb:


eddieleeper-nb:
[Fix bug in Needs List](https://trello.com/c/hLJc6zzT/301-fix-bug-in-needs-list)

PR #124, opened 2020-04-14
Feature/284 add e2e tests for updating a need

jamiebuckley:
Add acceptance tests to cover editing a need (changing status, assignee)
Allow running javascript dependent tests only in headless firefox

### DIFF
--- a/features/step_definitions/edit_need_steps.rb
+++ b/features/step_definitions/edit_need_steps.rb
@@ -4,20 +4,20 @@ Given('a resident with a need exists') do
 end
 
 When('I (assign)/(have assigned) the need to me') do
-  visit "needs/#{@need.id}"
+  visit "/needs/#{@need.id}"
   page.select 'admin@test.com', from: 'need_user_id'
   @expected_assignee = 'admin@test.com'
 end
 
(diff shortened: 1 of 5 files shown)


features/step_definitions/edit_need_steps.rb
```diff
@@ -0,0 +1,61 @@
+Given('a resident with a need exists') do
+  @contact = Contact.create!(first_name: 'Test')
+  @need = Need.create!(contact_id: @contact.id, name: 'Phone Triage', category: 'Phone Triage')
+end
+
+When('I (assign)/(have assigned) the need to me') do
+  visit "/needs/#{@need.id}"
+  page.select 'admin@test.com', from: 'need_user_id'
```
go8soft, line 8: In this case we expect Firefox to trigger js submit event. To have ci/cd test & builds we need Firefox and X on our dev machine.

jamiebuckley: So curiously firefox is already installed on the ubuntu-latest container used for the github actions:
https://github.com/wearefuturegov/beacon/runs/586438542?check_suite_focus=true

PR #125, opened 2020-04-15
hide users remove button for now

rik86future:


rik86future:
[BUG: Delete user breaks - Hide Link for now](https://trello.com/c/Y4r9y0Jb/309-bug-delete-user-breaks-hide-link-for-now)

timwis:
I think just delete the line entirely rather than leave commented out code in there. Can always restore from version control.

rik86future:
> I think just delete the line entirely rather than leave commented out code in there. Can always restore from version control.

Done

PR #126, opened 2020-04-15
add fields to needs export

rik86future:
Added the Covid19 question, amended the csv export and added a e2e test

image

image

image

### DIFF
--- a/app/views/contacts/_extra_questions.html.erb
+++ b/app/views/contacts/_extra_questions.html.erb
@@ -25,11 +25,11 @@
   <legend><h2 class="field-section__title">Are you or anyone in your home showing any symptoms of COVID-19?</h2></legend>
   <br/>
   <div class="radio">
-    <%= form.radio_button :has_covid, true, :id => "has_covid_true" %>
-    <label for="has_covid_true" class="radio__label">Yes</label>
+    <%= form.radio_button :has_covid_symptoms, true, :id => "has_covid_symptoms_true" %>
+    <label for="has_covid_symptoms_true" class="radio__label">Yes</label>
   </div>
   <div class="radio">
-    <%= form.radio_button :has_covid, false, :id => "has_covid_false" %>
-    <label for="has_covid_false" class="radio__label">No</label>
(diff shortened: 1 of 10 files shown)


rik86future:
[Add fields to needs export](https://trello.com/c/L7M9eGVU/304-add-fields-to-needs-export)

rik86future:
[Add fields to contact model](https://trello.com/c/UoGJRmZv/303-add-fields-to-contact-model)

app/views/shared/_profile-accordion.html.erb
```diff
@@ -1,6 +1,6 @@
 <section class="accordion">
 
-    <button class="accordion__control" aria-expanded="true">
+    <button class="accordion__control" aria-expanded="false">
```
timwis: Why did you change this? This block is expanded by default, so I think this should remain `true`, no?

db/schema.rb
```diff
@@ -58,6 +58,7 @@
     t.jsonb "healthintent_import_data"
     t.string "nhs_number"
     t.date "date_of_birth"
+    t.boolean "has_covid"
```
timwis, line 61: Can we change the field name to `has_covid_symptoms`? I think there may be a big difference and easily misinterpreted.

app/views/contacts/_extra_questions.html.erb
```diff
@@ -21,4 +21,15 @@
     <%= form.radio_button :any_children_below_15, false, :id => "any_children_below_15_false" %>
     <label for="any_children_below_15_false" class="radio__label">No</label>
   </div>
+
+  <legend><h2 class="field-section__title">Are you or anyone in your home showing any symptoms of COVID-19?</h2></legend>
```
timwis, line 25: These 3 questions should actually be wrapped in their own <fieldset> tags. that's why the display is a bit funky. Not a huge deal though.

app/views/shared/_profile-accordion.html.erb
eddieleeper-nb: I agree - this shouldn't change.

PR #127, opened 2020-04-15
IE11 accordion expansion/closing doesn't work

slav-petkov-saorsa:
Background:
https://trello.com/c/Js7GZBxK/307-ie-11-bug-accordions-not-closing

Solution:
- As it turned out, IE11 **doesn't support forEach** method on NodeList, which is retuend by **querySelectorAll**. Converting the NodeList to Array solves the problem.
- Expanded all accordions when loading the contacts view

### DIFF
--- a/app/assets/stylesheets/accordion.scss
+++ b/app/assets/stylesheets/accordion.scss
@@ -1,6 +1,11 @@
 @import "variables";
 
 .accordion{
+    div {
+      .accordion__content {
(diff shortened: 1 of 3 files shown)


slav-petkov-saorsa:
[IE 11 BUG: accordions not closing](https://trello.com/c/Js7GZBxK/307-ie-11-bug-accordions-not-closing)

PR #130, opened 2020-04-15
Feature/310 add scheduled call back functionality

jamiebuckley:


app/models/need.rb
```diff
@@ -120,4 +121,14 @@ def assigned
   def last_phoned_date
     read_attribute('last_phoned_date')
   end
+
+  # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+  def self.sort_created_and_start_date(first, second)
```
timwis, line 133: Maybe add a comment summarising what this does? e.g. "Sorts by start_on, then created_at" or something

app/views/contacts/_needs-table.html.erb
```diff
@@ -23,6 +23,14 @@
                 </td>
                 <td><%= n.assigned %></td>
                 <td><%= n.created_at&.strftime("%-d %B %Y") %></td>
+                <td>
+                  <%= n.start_on&.strftime("%-d %B %Y") %>
+
+                  <%# is this a 'future' need %>
+                  <% if !n.start_on.nil? && n.start_on > DateTime.now %>
+                     <strong>Padlock</strong>
```
timwis, line 31: In case we don't get round to adding the svg today, maybe write the word "Future" or something that would make sense for the time being

app/views/contacts/_needs-table.html.erb
```diff
@@ -23,6 +23,14 @@
                 </td>
                 <td><%= n.assigned %></td>
                 <td><%= n.created_at&.strftime("%-d %B %Y") %></td>
+                <td>
```
timwis, line 26: I think you're missing a `<th>` in the header to correspond to this.

app/views/needs/show.html.erb
```diff
@@ -33,6 +33,16 @@
           <dt class="details-list__caption">Description</dt>
           <dd class="details-list__value"><%= @need.name %></dd>
 
+          <% if @need.food_priority %>
```
timwis, line 36: We should also show start date on here, no?

jamiebuckley:
@timwis about displaying on the needs index page.  I agree, and we also probably need to highlight that a need is a future need there as well.  I'd rather do that in a separate PR though, as I suspect it will require reworking the cucumber tests, which need to be beefed up to cover this functionality too

timwis:
Don't wait on me to merge - I added an approval before, not sure if it's still there. If we need to add the field to the index page can you create a trello card in prioritised for it though?

jamiebuckley:
@timwis will do

PR #133, opened 2020-04-15
add radio for any dietary requirement question back

rik86future:
image

image

image


rik86future:
[Add dietary questions boolean back to UI and export](https://trello.com/c/K2IxZKap/319-add-dietary-questions-boolean-back-to-ui-and-export)

PR #134, opened 2020-04-15
Remove healthintent data

timwis:
The latest camden data does not include this anymore

PR #135, opened 2020-04-15
Add fields to needs export

timwis:
This is based on the list of fields the food delivery team expects. I switched to using a hash so we could control the field names.

(Note this is branched off of #134)


timwis:
[Confirm the needs export has at least all the fields the delivery team expects](https://trello.com/c/SrSbDQ77/315-confirm-the-needs-export-has-at-least-all-the-fields-the-delivery-team-expects)

PR #136, opened 2020-04-15
Remove validation from count_people_in_house

timwis:
I expect managers may be filling out this form without always having this infromation to hand. This validation made it so you can't submit the form without a number.


timwis:
[Don't require # of ppl in household](https://trello.com/c/mHc1aHq9/327-dont-require-of-ppl-in-household)

PR #139, opened 2020-04-15
Add seeds for new fields

timwis:
contacts.has_covid_symptoms, needs.food_priority, and needs.food_service_type


timwis:
[Add seeds for the new fields we've created](https://trello.com/c/5hLGt1Rg/314-add-seeds-for-the-new-fields-weve-created)

PR #140, opened 2020-04-16
Ensure Needs with no start date built via factorybot have nil start_on

jamiebuckley:
Explanation:

'Started' needs were built like this
```
let!(:created_tomorrow_started) { build :need, name: 'created tomorrow, no start date', created_at: DateTime.now + 1.days }
```
Which meant that whether or not those needs had a start date was decided by factorybot via:

```
start_on { [nil, Faker::Date.between(from: 2.days.ago, to: 6.days.from_now)].sample }
```

### DIFF
--- a/spec/models/need_spec.rb
+++ b/spec/models/need_spec.rb
@@ -37,8 +37,8 @@ RSpec.describe Need, type: :model do
   end
 
   describe 'sort_created_and_start_date' do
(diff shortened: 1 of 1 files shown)


jamiebuckley:
[Needs start date test occasionally fails](https://trello.com/c/n4T5PIWA/331-needs-start-date-test-occasionally-fails)

PR #142, opened 2020-04-16
Fix sorting on notes template

rik86future:
image


rik86future:
[Notes list sorts in reverse order](https://trello.com/c/yJS02VYN/326-notes-list-sorts-in-reverse-order)

PR #143, opened 2020-04-16
IE 11 BUG: Blank footer filling the page

slav-petkov-saorsa:
Background:
https://trello.com/c/UjosBuwP/338-ie-11-bug-blank-footer-filling-the-page

Solution:
- IE 10 & 11 have a bug with the min-height for a flex container. The workaround is to make the container a flex element itself, Therefore made the html a flex container as well,


rik86future:
I verified this fix the footer issue in IE11 
image

PR #123, opened 2020-04-14
fix bug with contact page validation

rik86future:
image

Steps to reproduce:
1) Open a contact
2) Click Triage call
3) Remove first name
4) Click save

PR #144, opened 2020-04-16
Feature/336 add basic triage form validation

jamiebuckley:
Triage form validation is complex with the previous implementation using need indexes when constructing the forms.

The reason behind the original implementation, was to be able to submit an array of hashes, some of which have fields edited via radio buttons.
This means that rails array syntax of `array_field[][value]` wouldn't work out of the box.

However, once validation is required, having fields that look like:
`f.text_field "need[#{index}]"` becomes complex, as validation requires posting and then repopulating values on a page if there is a validation error

This implementation aims to replace the indexes by wrapping each need with a `fields_for`, which should mean that each need form section won't need to concern itself with the array indexes


rik86future:
@jamiebuckley I have updated this branch with master but it's still failing some checks

jamiebuckley:
Thanks @rik86future, I expected that, working through some cucumber failures now

jamiebuckley:
Oh dear :( I'll take a look

rik86future:
Another one if you remove the placeholder in the date and click save 

image

jamiebuckley:
@rik86future this should improve things now

rik86future:
@jamiebuckley if I remove the first name, now the validation message does not appear at all. see https://github.com/wearefuturegov/beacon/pull/123

PR #145, opened 2020-04-17
Fix Pipeline

rik86future:
Fix the failing pipeline with error "ruby 2.5.7 not found"
After investigating I found the https://github.com/ruby/setup-ruby setup-ruby action was not supporting 2.5.7 hence I upgraded the whole project to 2.5.8 in order to minimise risks.

If you use `rbenv` you will need to fix your local setup as follows:
`rbenv install 2.5.8`
`rbenv global 2.5.8`
`gem install bundler`
`bundle install`


rik86future:
[Fix GitHub pipeline](https://trello.com/c/pItduyTH/339-fix-github-pipeline)

.github/workflows/run_cucumber_tests.yml
```diff
@@ -26,10 +26,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Ruby 2.5.7
+    - name: Set up Ruby 2.5.8
       uses: actions/setup-ruby@v1
       with:
-        ruby-version: 2.5.7
+        ruby-version: 2.5.8
```
jamiebuckley, line 32: So it looks like specifying the patch version doesn't do anything/is ignored when using this action, so maybe this should just be '2.5' or we use a different action that allows specifying the patch version

PR #148, opened 2020-04-20
Feature/340 spike role based access

jamiebuckley:


app/controllers/passwordless/sessions_controller.rb
```diff
@@ -48,6 +48,9 @@ def show
       BCrypt::Password.create(params[:token])
       sign_in passwordless_session
 
+      current_user.role = current_user.roles.first
+      current_user.save
```
timwis, line 52: Hm, not sure why you need to change anything in this controller, let alone persist it?

jamiebuckley: The user needs two forms of role:
1) A collection of roles they can adopt
2) A current role that they are viewing the system in

2 has to be persisted somewhere.  Storing it in the database was a simpler alternative to putting it in the session

jamiebuckley: The purpose of this bit, is when the user logs in, to find them a valid role they can be in (if they haven't picked one yet).
This part was kind of a temporary fix

app/models/user.rb
```diff
@@ -33,4 +36,20 @@ def name_or_email
   def last_logged_in
     passwordless_sessions.try(:last).try(:claimed_at)
   end
+
+  def role_type
+    role&.role
```
jamiebuckley, line 41: Not happy with role&.role.  Not sure what to name it exactly: the idea is that a Role has a friendly name, and than an internal identifier we can leave unchanged

jamiebuckley: Normally I'd suggest using a non-integer primary key, but it looks like [rails doesn't like that](https://stackoverflow.com/questions/1200568/using-rails-how-can-i-set-my-primary-key-to-not-be-an-integer-typed-column)

lib/tasks/db.rake
```diff
@@ -0,0 +1,14 @@
+namespace :db do
```
jamiebuckley, line 1: Simple rake task adding:
`rails db:drop:connections` to make dropping and recreating a postgres database easier

app/controllers/needs_controller.rb
```diff
@@ -10,8 +10,9 @@ class NeedsController < ApplicationController
   def index
     @params = params.permit(:user_id, :status, :category, :page, :order_dir, :order, :commit, :is_urgent)
     @users = User.all
```
eddieleeper-nb, line 12: Thinking about future requirements - this User.all will probably need to be reviewed.  i.e. should anyone be able to assign to anyone else, or should it be a restricted set.

app/controllers/passwordless/sessions_controller.rb
eddieleeper-nb: This currently breaks by default on my machine.  I think that there will need to be a database migration to give people roles initially, or just default to a minimum role if one isn't assigned.

app/controllers/passwordless/sessions_controller.rb
```diff
@@ -65,6 +70,14 @@ def destroy
       redirect_to passwordless_sign_out_redirect_path
     end
 
+    def set_role
```
eddieleeper-nb, line 73: Need to ensure that the role ids never change.

app/controllers/needs_controller.rb
jamiebuckley: Possibly true.  I don't think there's a requirement at the moment for who you can assign it to, but it does open the possibility of assigning to a user accidentally that you then can't see, e.g. a council service team assigning to a call centre user/team will no longer be able to see the need.

This can be solved currently via an admin user however

app/controllers/passwordless/sessions_controller.rb
jamiebuckley: How does this break?  Is the exception thrown or does it break after that?

app/controllers/passwordless/sessions_controller.rb
jamiebuckley: True.  Any idea on a better way to do this?

app/controllers/passwordless/sessions_controller.rb
jamiebuckley: Rails doesn't like data migrations, could maybe add a rake task to do this?  I wonder if a one off script would be better though

app/controllers/users_controller.rb
```diff
@@ -1,52 +1,77 @@
 # frozen_string_literal: true
 
 class UsersController < ApplicationController
-  before_action :set_user, only: [:destroy]
-  before_action :require_admin
+  before_action :set_user, only: [:destroy, :edit, :update]
 
   # GET /users
   def index
+    authorize User
```
eddieleeper-nb, line 8: For all the authorize checks - if the user isn't authorize then we'll get an application exception at the moment I think (I'll try  it out shortly to check) - I  guess we might just want to redirect to the home page with an unauthorized flash message.

app/controllers/passwordless/sessions_controller.rb
jamiebuckley: Actually I don't think this is a problem.  RoleIDs shouldn't change - if there is a requirement to disable roles later, that should be a separate concern.

app/controllers/users_controller.rb
```diff
@@ -1,52 +1,77 @@
...
     @users = User.all.page(params[:page])
   end
 
   # GET /users/new
   def new
+    authorize User
```
eddieleeper-nb, line 14: maybe use a before_action where required to `authorize User`

jamiebuckley: https://stackoverflow.com/questions/23983576/why-does-before-action-authorize-fail-with-wrong-number-of-arguments suggests that before_actions shouldn't be used with pundit.  It seems kind of copy-paste to me too, but pundit in general seems to encourage scattering the authorize calls in the controller methods

app/controllers/users_controller.rb
jamiebuckley: Yep good point, will change that

app/models/user.rb
```diff
@@ -12,13 +12,16 @@ class User < ApplicationRecord
   has_many :uncompleted_needs, -> { uncompleted }, class_name: 'Need'
   has_many :uncompleted_contacts, through: :uncompleted_needs, source: :contact
   has_many :completed_contacts, through: :completed_needs, source: :contact
+  has_many :user_roles, dependent: :destroy
+  has_many :roles, through: :user_roles
+  belongs_to :role, optional: true
```
eddieleeper-nb, line 17: I'm not sure that the role should be optional.

app/models/user_role.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class UserRole < ApplicationRecord
+  belongs_to :user
```
eddieleeper-nb, line 4: You could consider using Polymorphic associations here.  Although I think all the models could do with a thorough review -if we had more than a few days left!

app/policies/application_policy.rb
```diff
@@ -0,0 +1,55 @@
+class ApplicationPolicy
```
eddieleeper-nb, line 1: I'll skim over the policies - as I think that's one of the key areas that you will be working on in the next step for this feature.

app/policies/user_policy.rb
```diff
@@ -0,0 +1,5 @@
+class UserPolicy < ApplicationPolicy
+  [:index?, :update?, :create?].each do |m|
+    define_method(m) { admin? }
```
eddieleeper-nb, line 3: fancy 😀

app/views/layouts/application.html.erb
```diff
@@ -10,6 +10,20 @@
       <header class="panel content-header">
         <%= yield :header %>
         <nav class="content-header__actions-right">
+
+          <div class="content-header__roles">
```
eddieleeper-nb, line 14: I like this - ideally it would be nice to have a non-JS fallback implementation, but we've also got other JS dependencies that we don't have time to address.

app/controllers/users_controller.rb
jamiebuckley: This now redirects to root on any pundit error, see application_controller

app/models/user.rb
jamiebuckley: This links in with the data migration issue, e.g. need to add this field, and also assign to a role at the same time.

app/models/user_role.rb
jamiebuckley: This is a join table rather than an either-or relationship. That's a new thing to me in rails though, nice feature.  Worth suggesting for the 'needs can be assigned to a user or a role' requirement!

app/policies/application_policy.rb
jamiebuckley: Yep, at the moment the only ones that really do anything are the admin/agent ones, which behave roughly the same as they do currently, e.g. only admins can create users.  Nothing else really does anything

app/views/layouts/application.html.erb
jamiebuckley: Actually it functions without JS :smile: if you turn off JS it's a dropdown and a 'switch' submit button.  The JS removes the button and makes it autosubmit

jamiebuckley:
[Role based access](https://trello.com/c/Ob0XyF6z/340-role-based-access)

app/controllers/passwordless/sessions_controller.rb
eddieleeper-nb: Sorry - It might have just been the first time in. It's working now - looking for the screenshot.  I'm wondering if it should throw an exception here?  like Tim said the passwordless session controller should probably be as slim and minimal as possible - it doesn't need to know about roles.  The other controllers have the policies and authorize calls - it feels like any role behaviour belongs there.

app/views/layouts/application.html.erb
eddieleeper-nb: That's what I get for commenting too early - without going throught the whole code :-)

config/routes.rb
```diff
@@ -15,6 +15,7 @@
     resources :notes
   end
 
-  resources :users, only: [:index, :new, :create, :destroy]
+  resources :users, only: [:index, :new, :create, :edit, :update, :destroy]
+  post 'role', to: 'passwordless/sessions#set_role', as: 'set_role'
```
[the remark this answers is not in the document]
jamiebuckley: I was thinking of it more as the 'sessions_controller' e.g. controlling your session, but I can move it if you thinks that's better

db/seeds.rb
```diff
@@ -40,15 +40,27 @@
     end
   end
 
+  {
+    'Contact Centre Manager' => 'manager',
```
eddieleeper-nb, line 44: The roles need to be available in production/staging etc.  I don't think we should run the seeds.rb file in production.

features/invite_user.feature
```diff
@@ -13,26 +13,10 @@ Feature: Invite user
     Then I see a user created message
     And the email address is in the list of users
```
eddieleeper-nb, line 15: I think it would be good to get these tests - and additional tests passing - as this was initially a spike I guess you plan to update and add in the next trello task.

app/views/shared/_side-nav.html.erb
```diff
@@ -40,13 +40,13 @@
...
         </li>
 
-        <% if current_user.admin %>
-            <li class="main-menu__item">
-                <%= link_to_unless_current "Users", users_path, class: "main-menu__link" do %>
-                    <span class="main-menu__current">Users</span>
-                <% end %>
-            </li>
-        <% end %>
+      <% if policy(User).index? %>
```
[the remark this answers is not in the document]
jamiebuckley: Not sure I understand this, do you mean the code or the general approach doesn't feel right?

Taken from pundit docs below:
```
<% if policy(:dashboard).show? %>
  <%= link_to 'Dashboard', dashboard_path %>
<% end %>
```

db/seeds.rb
jamiebuckley: Yep, definitely, there's a need for a data migration scripts to create and assign these roles, will add a trello card for that

features/invite_user.feature
jamiebuckley: Yep, didn't want to write lots of tests when I wasn't sure of the implementation. Was going to add those when I implement all of the roles

app/controllers/passwordless/sessions_controller.rb
jamiebuckley: Looking at removing this code - it will raise an exception later when a user has a null role, are we ok with that as an alternative?

jamiebuckley: @eddieleeper-nb I have an alternative way of doing this now, outside the passwordless controller, better?

PR #151, opened 2020-04-22
358 14 inbound enquiry capture 2 data retention

rik86future:
image

image

### DIFF
--- a/db/migrate/20200422121348_add_channel_to_contacts.rb
+++ b/db/migrate/20200422121348_add_channel_to_contacts.rb
@@ -1,0 +1,5 @@
+class AddChannelToContacts < ActiveRecord::Migration[6.0]
+  def change
+    add_column :contacts, :channel, :string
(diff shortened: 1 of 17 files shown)


db/migrate/20200422121348_add_channel_to_contacts.rb
```diff
@@ -0,0 +1,5 @@
+class AddChannelToContacts < ActiveRecord::Migration[6.0]
+  def change
+    add_column :contacts, :channel, :string
```
eddieleeper-nb, line 3: I wonder if we should have a default here - i.e. `bulk import`

rik86future: yes good one

rik86future: or shall we call it "shield import"?

PR #155, opened 2020-04-23
add js fix

rik86future:


app/javascript/packs/users_viewing_page.js
```diff
@@ -32,6 +32,13 @@ const subscription = consumer.subscriptions.create({ channel: 'ContactChannel',
 });
 
 
+const cleanup = () => {
```
rik86future, line 35: it must have disappeared in the merge

timwis: Can we remove the `console.log()` before merging?

rik86future: sure

PR #158, opened 2020-04-23
Feature/361 2 restriction on what data users can see

jamiebuckley:


app/controllers/contacts_controller.rb
```diff
@@ -3,6 +3,7 @@
 class ContactsController < ApplicationController
   before_action :set_contact, only: %i[edit update show needs add_needs]
 
+  # to do - search with policy scope?
```
jamiebuckley, line 6: I haven't done much work into checking if the roles work with the search functionality as it was merged later after I had started, will look into this now though

app/models/role.rb
```diff
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class Role < ApplicationRecord
+  alias_attribute :tag, :role
```
jamiebuckley, line 4: The reason for this is that I chose a poor name for the 'role' field on the role, i.e. the internal identifier for the role.  In some queries, seeing `role.role` made it difficult to understand.  I can change this properly if it's considered necessary

app/controllers/contacts_controller.rb
eddieleeper-nb: I've asked in the dev channel.  I think we would restrict roles to either be able to search or not.  We don't have the time to be more fine grained at the moment.

jamiebuckley: The PR has been updated to make search operate on scoped contacts

app/controllers/needs_controller.rb
```diff
@@ -47,6 +46,7 @@ def update_multiple
     for_update = JSON.parse(params[:for_update])
     for_update.each do |obj|
       need = Need.find(obj['need_id'])
+      authorize(@need)
```
eddieleeper-nb, line 49: This doesn't look right - should be `need`

jamiebuckley: Good spot, this is a mistake.  Fixed

PR #159, opened 2020-04-23
Drop unused tables

timwis:
Removes contact lists and organisations, and ActsAsTaggableOn. None of these were in use since the early days of the project, but the database tables still existed.

The data model is now just these tables:

data-model-2020-04-23


eddieleeper-nb:
[Remove organisations, contact lists, and tags](https://trello.com/c/TzdwKMRy/294-remove-organisations-contact-lists-and-tags)

PR #160, opened 2020-04-24
377 add date of birth and nhs number to the profile

eddieleeper-nb:


eddieleeper-nb:
[Display new DOB and NHS number fields](https://trello.com/c/5z4TxhGO/377-display-new-dob-and-nhs-number-fields)

PR #161, opened 2020-04-24
Feature/add need status

go8soft:
Add `need`.`status` regarding task:
https://trello.com/c/tVy0goVC/374-13-blocked-need-status
Please also check filters, export and need navs (Assigned to me and Unmet needs). 
Please take a note we have fallback need label "To do" when the new status column is null. We can not filter by it.

https://github.com/wearefuturegov/beacon/pull/161/commits/947306db484c68e444657307d9381d9430714398 is a rake task. Rake tasks are one time task running in Rails project environment. In our case it means we dont care about database configurations because will come from Rials env. To run this rake task:

```
i-have-i-need$ bundle exec rake needs_status:update_needs_statuses
```


spec/models/need_spec.rb
```diff
@@ -14,8 +14,8 @@
   end
 
   describe 'scopes' do
-    let!(:uncompleted_need) { create :need }
-    let!(:completed_need) { create :need, :completed }
+    let!(:uncompleted_need) { create :need, status: :to_do }
```
eddieleeper-nb, line 17: Should this actually be any state that in't complete?

spec/models/need_spec.rb
```diff
@@ -29,11 +29,14 @@
   it { is_expected.to be_versioned }
 
   it '#status' do
-    need1 = build :need, name: 'medicines', completed_on: nil
-    expect(need1.status).to eq 'To do'
+    need1 = build :need, name: 'medicines', status: :to_do
```
eddieleeper-nb, line 32: I guess there should be tests for the other states too.

spec/models/user_spec.rb
```diff
@@ -11,8 +11,8 @@
     it { is_expected.to have_many(:needs).dependent(:destroy) }
     it { is_expected.to have_many(:notes).dependent(:destroy) }
     it { is_expected.to have_many(:assigned_contacts).through(:needs).source(:contact) }
-    it { is_expected.to have_many(:uncompleted_needs).conditions(completed_on: nil) }
-    it { is_expected.to have_many(:completed_needs).conditions('completed_on IS NOT NULL') }
+    it { is_expected.to have_many(:uncompleted_needs).conditions('status <> complete') }
```
eddieleeper-nb, line 14: see comment above - I think this is correct - above needs updating.

easyauthn: Not sure what you mean. Here we are testing needs sopes completed and uncompleted.

go8soft: Not sure what you mean. Here we are testing needs sopes completed and uncompleted.

jamiebuckley:
Hmm, seems to work for me

go8soft:
> The unmet needs filter seems to be broken.
> To recreate click unmet needs on the left - I get `no matches`
> Or update the filters manually.

Please take a note we have fallback need label "To do" when the new status column is null. We can not filter by it. Is that the case?

spec/models/need_spec.rb
go8soft: Yes, and we start from `to_do`. This is why I put this one.

spec/models/need_spec.rb
go8soft: Thanks, done.

eddieleeper-nb:
[13. Blocked “Need Status”](https://trello.com/c/tVy0goVC/374-13-blocked-need-status)

go8soft:
Need to improve the rake before merge to master.

go8soft:
Fix to ensure if we run rake task twice:
https://github.com/wearefuturegov/beacon/pull/161/commits/be211966a6e66400fa7a4ad9f8051ac6921a34b3

db/migrate/20200328094815_acts_as_taggable_on_migration.acts_as_taggable_on_engine.rb
```diff
@@ -1,13 +1,13 @@
 # This migration comes from acts_as_taggable_on_engine (originally 1)
 class ActsAsTaggableOnMigration < ActiveRecord::Migration[6.0]
   def self.up
-    create_table ActsAsTaggableOn.tags_table do |t|
```
jamiebuckley, line 4: Should all of these old migrations have been changed?

jamiebuckley: Ah, I suppose since we are removing the gem, they have to be.  Will this work fine on production then?

go8soft: @jamiebuckley this is @timwis changes.

PR #167, opened 2020-04-26
Refactor: Use enum for need category

timwis:
This refactor uses an enum for `category` in the need model, which simplifies working with the list of categories a bit.

Significantly, it moves the border/stroke colours to CSS. This involves hard-coding categories in CSS classes. I considered simply having a CSS variable that was a list of colours, and the stroke colour would be based on the index in the enum. That would probably be better but might be a bit less obvious what's going on, so this seemed fine. I could go either way though.

I've got some failing cucumber tests, and I _think_ it may be because of some of my refactoring in the triage controller. In particular, I'm not sure what `merge_contact_needs` was meant to do or if we still need it now that colours aren't involved in the list. @jamiebuckley can you take a peek when you have a free minute?

This is not an urgent feature — and I'm fine with discarding it — but my hope is it may make it slightly easier to implement some of the other features we need to do, and also demonstrates how to fix the filtering bug on the status column (cc @go8soft).

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,3 +82,4 @@ Coming soon...
 | NOTIFY_API_KEY   | Gov.uk Notify API key (production only)   |
 | HOSTNAME         | Hostname used in outbound emails (e.g. `x.beacon.com`). Defaults to heroku app name |
 | SEED_USER_EMAILS | Optional comma-separated list of emails to seed users table with |
+
--- a/app/assets/stylesheets/_layout.scss
+++ b/app/assets/stylesheets/_layout.scss
@@ -27,6 +27,10 @@ body{
     &.panel--unpadded{
         padding: 0px;
     }
+    &.panel--stroked {
+        border-left-width: 5px;
+        border-left-style: solid;
+    }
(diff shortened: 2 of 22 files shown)


go8soft:
Thanks @timwis, what is the filtering bug on status column you are talking about?

jamiebuckley:
Vast majority of the test failures look to be a mismatch between 'Phone triage' and 'Phone Triage', e.g. just the case difference.

An explanation of merge_contact_needs:
The helper (which was just a method returning labels and CSS mappings) was the only real part of the code that defined all the needs.  The triage form would - when doing a new triage, use this helper to generate the list of forms items to present the user.

When a user submitted a triage form and got a validation error - or when the session repopulates the form - this info (label and css field) would be lost, and so would need to be merged back in.

jamiebuckley:
[Replacing split needs triage form/needs helper with needs enum](https://trello.com/c/qCFplGOt/418-replacing-split-needs-triage-form-needs-helper-with-needs-enum)

PR #171, opened 2020-04-27
375 10 allow assigning a need to a role pt2

rik86future:
Supposed to be looking like this:

image

Actual:

image

Also tested on IE11


rik86future:
[10. Allow assigning a need to a role (pt2)](https://trello.com/c/E1Ice7Yz/375-10-allow-assigning-a-need-to-a-role-pt2)

jamiebuckley:
This looks good.  Hopefully @slav-petkov-saorsa can have a look at the styling issues.
One thing - I'm not sure if we need to unassign users when assigning a team, and vice versa

PR #180, opened 2020-04-28
add check in need

rik86future:


rik86future:
[Add Check-in need type that’s hidden from the list of needs in the triage form](https://trello.com/c/Sl6w0WT3/422-add-check-in-need-type-thats-hidden-from-the-list-of-needs-in-the-triage-form)

timwis:
> that’s hidden from the list of needs in the triage form

😬 i think this just needs to be an addition to the enum, is all

PR #181, opened 2020-04-28
Warn users if they’re leaving the create/edit contact page with unsaved changes

slav-petkov-saorsa:
Background:
https://trello.com/c/3BZgNJez/428-warn-users-if-theyre-leaving-the-create-edit-contact-page-with-unsaved-changes

Solution: 
- I use the "onbeforeunload" event. Since rails turbo links prevent full page reload, I disable them only on the page where the warning is needed. After we leave the page, they are restored.

### DIFF
--- a/app/javascript/packs/contact-form-navigation-alert.js
+++ b/app/javascript/packs/contact-form-navigation-alert.js
@@ -1,0 +1,23 @@
+$(() => {
+   $("body").attr("data-turbolinks", false);
+
+   const formElement = $("form");
+   const originalState = formElement.serialize();
+   formElement.on("submit", onSubmit)
+
+   function shouldNavigateDirectly() {
+      const currentState = $("form").serialize();
+
+      $("*").blur();
+      return currentState === originalState ? undefined : false;
+   };
+
+   function onSubmit() {
+      $(window).off("beforeunload", shouldNavigateDirectly);
+   }
+
+   $(window).on("beforeunload", shouldNavigateDirectly);
+})
+
+
+
--- a/app/javascript/packs/contact-form-navigation-alert.js
+++ b/app/javascript/packs/contact-form-navigation-alert.js
@@ -1,7 +1,7 @@
 $(() => {
    $("body").attr("data-turbolinks", false);
 
-   const formElement = $("form");
+   const formElement = $("form.contact-form");
    const originalState = formElement.serialize();
    formElement.on("submit", onSubmit)
 
--- a/app/views/contacts/_form.html.erb
+++ b/app/views/contacts/_form.html.erb
@@ -1,5 +1,5 @@
 <div class="two-thirds">
-  <%= form_with(model: contact, local: true) do |form| %>
+  <%= form_with(model: contact, local: true, html: { class: "contact-form" }) do |form| %>
     <%= render partial: 'form_fields', locals: { contact: @contact, form: form } %>
     <div class="actions">
       <%= form.button "Save changes", class: "button button--primary" %>
--- a/app/views/contacts/edit.html.erb
+++ b/app/views/contacts/edit.html.erb
@@ -9,3 +9,4 @@
 
 
 <%= javascript_pack_tag 'users_viewing_page.js' %>
+<%= javascript_pack_tag 'contact-form-navigation-alert.js' %>
--- a/app/views/contacts/new.html.erb
+++ b/app/views/contacts/new.html.erb
@@ -5,7 +5,7 @@
 
 <!-- New contact form -->
 <div class="panel">
-  <%= form_with(model: @contact, local: true, class: "margin-bottom-70 two-thirds") do |f| %>   
+  <%= form_with(model: @contact, local: true, class: "margin-bottom-70 two-thirds contact-form") do |f| %>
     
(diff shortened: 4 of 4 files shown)


slav-petkov-saorsa:
[Warn users if they’re leaving the create/edit contact page with unsaved changes](https://trello.com/c/3BZgNJez/428-warn-users-if-theyre-leaving-the-create-edit-contact-page-with-unsaved-changes)

app/javascript/packs/contact-form-navigation-alert.js
```diff
@@ -0,0 +1,23 @@
...
+
+   const formElement = $("form");
+   const originalState = formElement.serialize();
+   formElement.on("submit", onSubmit)
+
+   function shouldNavigateDirectly() {
+      const currentState = $("form").serialize();
+
+      $("*").blur();
+      return currentState === originalState ? undefined : false;
```
timwis, line 12: I don't think this would ever evaluate to `true` because `===` between objects checks if they are the same piece of memory, rather than comparing their values.

app/javascript/packs/contact-form-navigation-alert.js
```diff
@@ -0,0 +1,23 @@
+$(() => {
+   $("body").attr("data-turbolinks", false);
```
timwis, line 2: What's this for?

app/javascript/packs/contact-form-navigation-alert.js
```diff
@@ -0,0 +1,23 @@
...
+
+   const formElement = $("form");
```
timwis, line 4: May want to be more specific - if we have a `<form>` in the nav bar to switch roles, or something, it would catch that.

app/javascript/packs/contact-form-navigation-alert.js
slav-petkov-saorsa: The result of the serialization is a string. When we use === for strings in JavaScript, we compare value and type. So, if the strings are equal by value, this will return true.

Quick snippet from the console:
image

app/javascript/packs/contact-form-navigation-alert.js
slav-petkov-saorsa: Turbo links in rails change pages without causing a full page reload. As it turns out, the onbeforeunload works only with full page reloads. When setting data-turbolinks attribute to false, we disable the turbolink behavior only for the specific page. When we leave it this change will disappear.

app/javascript/packs/contact-form-navigation-alert.js
slav-petkov-saorsa: > Hm, I was under the impression this could be simpler. Something like:
> 
> ```js
> let contactForm = document.querySelector('form#edit-contact, form#create-contact')
> let isContactFormDirty = false
> 
> if (contactForm) {
>   contactForm.addEventListener('change', () => isContactFormDirty = true)
>   contactForm.addEventListener('submit', () => isContactFormDirty = false)
> }
> 
> window.addEventListener('beforeunload', (event) => {
>   if (isContactFormDirty) {
>     e.preventDefault()
>     e.returnValue = ''
>   }
> })
> ```
> 
> Would that work?

The issue with the change event is that it will consider the form dirty even if the user restores the original values. Serialization will take the restored original values into account and the confirmation dialog won't be displayed.

app/javascript/packs/contact-form-navigation-alert.js
timwis: oh! I assumed it serialised to an object

timwis, line 12: Have you tested this on multiple browsers? MDN docs suggest we need to do `e.preventDefault()` and `e.returnValue = ''` https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload#Browser_compatibility but maybe that's outdated

app/javascript/packs/contact-form-navigation-alert.js
```diff
@@ -0,0 +1,23 @@
...
+   const originalState = formElement.serialize();
+   formElement.on("submit", onSubmit)
+
+   function shouldNavigateDirectly() {
+      const currentState = $("form").serialize();
+
+      $("*").blur();
```
timwis, line 11: Just curious what this was for?

app/javascript/packs/contact-form-navigation-alert.js
timwis: Got it. Note, though, that my question in this comment was actually about the specificity of the `'form'` selector, which I still think would be a bit of an issue.

app/javascript/packs/contact-form-navigation-alert.js
slav-petkov-saorsa: Yes, I tested it on IE 11, Chrome, Firefox and Edge

app/javascript/packs/contact-form-navigation-alert.js
slav-petkov-saorsa: If you decide to stay on the page, the clicked element remained focused. So, I clear the focus of all elements.

PR #183, opened 2020-04-28
Feature/profile edit in accordion on triage

jamiebuckley:


app/javascript/packs/triage_datepicker.js
```diff
@@ -6,4 +6,10 @@ flatpickr(dateElement, {
     dateFormat: "d/m/Y",
     defaultDate: dateElement.value ? null : new Date().fp_incr(6),
     allowInput: true
+});
+
+const contactDobElement = document.getElementById("contact_date_of_birth");
+flatpickr(contactDobElement, {
```
timwis, line 12: Alternatively, you could do `flatpickr('.datepicker')`. I think we already set the default date server-side for the phone triage need anyway (which we probably won't end up doing anymore in the assessment flow anyway).

app/views/shared/_profile-accordion-edit.html.erb
```diff
@@ -0,0 +1,120 @@
+<section class="accordion">
+
+  <button class="accordion__control" aria-expanded="true">
+    <h2 class="accordion__title">Personal details</h2>
+  </button>
+  <div class="accordion__content">
+    <%= form.hidden_field :lock_version, :value => contact.lock_version %>
+    <div class="field">
```
timwis, line 8: I thought we were going to make each accordion section a partial so we could reuse them in the create contact form? More trouble than it's worth?

app/views/triage/edit.html.erb
```diff
@@ -99,11 +92,9 @@
...
-    <% end %>
   </aside>
 </div>
-
+<% end %>
 <%= javascript_pack_tag 'users_viewing_page.js' %>
-
-<%= javascript_pack_tag 'save-for-later.js' %>
\ No newline at end of file
+<%= javascript_pack_tag 'save-for-later.js' %>
```
timwis, line 99: I assume drafts would be removed in a separate PR?

app/views/shared/_profile-accordion-edit.html.erb
jamiebuckley: Yeah, this ended up being quite tricky as 
1: the styling and tags vary quite a bit between the accordion and the profile edit page, e.g. there are massive h2 elements in the edit profile page that serve as labels, which look out of place in the accordion
2: There are certain contact fields that are being edited alongside the need, that have to be excluded from the accordion

app/views/triage/edit.html.erb
jamiebuckley: Yep, thought it was best to scope each PR just to one thing

app/javascript/packs/triage_datepicker.js
jamiebuckley: true, that would be nicer.  There are some quirks with how I've built the phone triage need validation etc. once the phone triage has been removed from this 'add-needs' page will simplify

PR #184, opened 2020-04-28
Show the number of call attempts on the list of support actions

go8soft:
Task: https://trello.com/c/rVAGO3a7/415-show-the-number-of-call-attempts-on-the-list-of-support-actions


app/models/need.rb
```diff
@@ -151,15 +151,22 @@ def self.base_query
           group by c.id) as contact_aggregation
           on contact_aggregation.id = contacts.id"
 
-    Need.joins(:contact, sql).select('needs.*', 'contact_aggregation.max as last_phoned_date')
+    sql += " left join (
+          select n.id, count(nt.id) from notes nt
+          join needs n on n.id = nt.need_id and nt.category in ('phone_success', 'phone_message', 'phone_failure')
```
timwis, line 156: Wouldn't a more native rails approach be to use a scope in `Note` like `scope calls: { where.in('phone_success', ...) }` and then preload it in the `Need` model to avoid n+1?

go8soft: Actually we have only 3 requests with base_query per page:

1 - Need Exists?
```
SELECT 1 AS one FROM "needs" INNER JOIN "contacts" ON "contacts"."id" = "needs"."contact_id" LEFT JOIN (select c.id, max(nt.created_at) from contacts c
          left join needs n on n.contact_id = c.id
          left join notes nt on nt.need_id = n.id where nt.category like 'phone_%'
          group by c.id) as contact_aggregation
          on contact_aggregation.id = contacts.id left join (
          select n.id, count(nt.id) from notes nt
          join needs n on n.id = nt.need_id and nt.category in ('phone_success', 'phone_message', 'phone_failure')
          group by n.id
        ) as notes_aggr on notes_aggr.id = needs.id WHERE (start_on IS NULL or start_on <= '2020-04-28') LIMIT $1 OFFSET $2  [["LIMIT", 1], ["OFFSET", 0]]
  ↳ app/views/needs/index.html.erb:19
```

2 -   Need Load 
```
SELECT needs.*, contact_aggregation.max as last_phoned_date, notes_aggr.count as call_attempts  FROM "needs" INNER JOIN "contacts" ON "contacts"."id" = "needs"."contact_id" LEFT JOIN (select c.id, max(nt.created_at) from contacts c
          left join needs n on n.contact_id = c.id
          left join notes nt on nt.need_id = n.id where nt.category like 'phone_%'
          group by c.id) as contact_aggregation
          on contact_aggregation.id = contacts.id left join (
          select n.id, count(nt.id) from notes nt
          join needs n on n.id = nt.need_id and nt.category in ('phone_success', 'phone_message', 'phone_failure')
          group by n.id
        ) as notes_aggr on notes_aggr.id = needs.id WHERE (start_on IS NULL or start_on <= '2020-04-28') ORDER BY "needs"."created_at" ASC LIMIT $1 OFFSET $2  [["LIMIT", 10], ["OFFSET", 0]]
  ↳ app/views/needs/index.html.erb:73
```

3 - Paging counter
```
  SELECT COUNT(*) FROM "needs" INNER JOIN "contacts" ON "contacts"."id" = "needs"."contact_id" LEFT JOIN (select c.id, max(nt.created_at) from contacts c
          left join needs n on n.contact_id = c.id
          left join notes nt on nt.need_id = n.id where nt.category like 'phone_%'
          group by c.id) as contact_aggregation
          on contact_aggregation.id = contacts.id left join (
          select n.id, count(nt.id) from notes nt
          join needs n on n.id = nt.need_id and nt.category in ('phone_success', 'phone_message', 'phone_failure')
          group by n.id
        ) as notes_aggr on notes_aggr.id = needs.id WHERE (start_on IS NULL or start_on <= '2020-04-28')
  ↳ app/views/shared/_pagination.html.erb:3
```

We dont have n+1 problem in this case.

timwis:
I think @jamiebuckley recently went from raw SQL back to active record on here, so I'd like his eyes on this before merging, if we're going back to raw SQL if that's okay.

jamiebuckley:
I looked into it, it's actually quite hard to do count fields joined into a list of records without SQL, rails seems to only have advice on getting aggregate values on a single item.  This approach seems fine to me

PR #186, opened 2020-04-29
Style role-switcher drop down to match look and feel

slav-petkov-saorsa:
Background:
https://trello.com/c/TSl1ccgK/431-style-role-switcher-drop-down-to-match-look-and-feel

Solution:
- Introduced new class "dropdown" which can be applied on all menus.
- Decided to leave role label, because it doesn't make sense for the role menu to have a default option which prompts the user to select something.


slav-petkov-saorsa:
[Style role-switcher drop down to match look and feel](https://trello.com/c/TSl1ccgK/431-style-role-switcher-drop-down-to-match-look-and-feel)

jamiebuckley:
Screenshot from 2020-04-29 14-03-00
Looks good, but might be good to double the width, a lot of role names are quite long

slav-petkov-saorsa:
Actual layout:

image

PR #187, opened 2020-04-29
Remove needs counter_culture

go8soft:
Task: https://trello.com/c/dJS5FsI0/109-remove-cached-task-counts-from-contacts-model


go8soft:
I need to remove the columns, too

PR #188, opened 2020-04-29
add a new person

rik86future:


rik86future:
[Add Create contact page](https://trello.com/c/NzCRFAp1/424-add-create-contact-page)

config/initializers/time_formats.rb
```diff
@@ -0,0 +1,3 @@
+# Default format for displaying dates and times
+Date::DATE_FORMATS[:default] = '%d/%m/%Y'
```
timwis, line 2: What's this for? We actually use `%-d %B %Y` or something in most places, if I recall correctly. And shouldn't the Time one have a time in it?

rik86future: I will remove it

rik86future: it does who dates like this without it
image

rik86future: "%-d %B %Y" is used in plain prints, here I am doing a input text this is the main diff

PR #189, opened 2020-04-29
Feature/remove needs counter culture

go8soft:
Task: https://trello.com/c/dJS5FsI0/109-remove-cached-task-counts-from-contacts-model

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -2,6 +2,7 @@
 
 class Contact < ApplicationRecord
   include PgSearch::Model
+  self.ignored_columns = ['needs_count', 'uncompleted_needs_count', 'completed_needs_count']
 
   has_many :needs, dependent: :destroy
   has_many :uncompleted_needs, -> { uncompleted }, class_name: 'Need'
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -2,7 +2,6 @@
 
 class Contact < ApplicationRecord
   include PgSearch::Model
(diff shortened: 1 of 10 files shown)


app/models/contact.rb
```diff
@@ -2,6 +2,7 @@
 
 class Contact < ApplicationRecord
   include PgSearch::Model
+  self.ignored_columns = ['needs_count', 'uncompleted_needs_count', 'completed_needs_count']
```
timwis, line 5: Why do we need this if we've removed the columns?

db/seeds.rb
```diff
@@ -63,7 +63,7 @@
                             contact: contact,
                             user: [user, nil].sample,
                             is_urgent: [true, false].sample,
-                            completed_on: [nil, [1,2,3].sample.days.ago].sample
+                            status: ['to_do', 'complete'].sample
```
timwis, line 66: Shouldn't we use `Need.status.sample` instead so other statuses get included in the seeds?

app/models/contact.rb
go8soft: Not necessarily, just good practice. Get it out of here?

go8soft: Done.

db/seeds.rb
go8soft: Thanks

go8soft:
> Looks good, just left a couple comments that we might want to fix before merging
Thanks, I applied the changes.

PR #190, opened 2020-04-29
Feature/430 add assessment flow

jamiebuckley:


app/controllers/assessments_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  private
+
+  def set_contact
+    @contact = Contact.find(params[:contact_id])
+  end
+
+  def log_assessment
+    @need = Need.new(assessment_params.merge(contact_id: @contact.id, name: 'Logged Assessment'))
+    @note = Note.new(notes_params.merge(need: @need))
```
timwis, line 33: We want to mark this category as `phone_success`, right?

timwis, line 33: Interesting: the `need_id` of the note must come from a hidden html element or something, as opposed to associating them here? 🤔

app/controllers/assessments_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+    @type = %w[log schedule].include?(type_param) ? type_param : 'log'
+
+    @need = Need.new
+    @note = Note.new
+
+    @need.status = 'complete' if @type == 'log'
+    @need.user = current_user if @type == 'log'
+  end
+
+  def create
```
timwis, line 16: Do we need to `authorize()` anything here?

app/controllers/assessments_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+      return
+    end
+
+    @users = User.all
+    @roles = Role.all
+    render :new
+  end
+
+  def schedule_assessment
+    @need = Need.new(assessment_params.merge(contact_id: @contact.id, name: 'Scheduled Assessment'))
```
timwis, line 45: Hm, I think when scheduling an assessment, we want the user to be able to provide a description (e.g. "This person needs a check-in to see if they still need food delivered")

app/javascript/packs/assessments.js
```diff
@@ -0,0 +1,10 @@
+document.querySelectorAll('.need--status_complete').forEach(e => e.style.display = 'none');
+
+let showCompletedNeeds = false;
+const toggleDisplay = (show) => show ? 'table-row' : 'none';
+
+document.getElementById('toggle-visibility-completed-assessments').addEventListener('click', (e) => {
```
timwis, line 6: JS looks great

app/models/need.rb
```diff
@@ -39,6 +42,9 @@ class Need < ApplicationRecord
   scope :started, -> { where('start_on IS NULL or start_on <= ?', Date.today) }
   scope :filter_by_category, ->(category) { where(category: category.downcase) }
 
+  scope :assessments, -> { where(category: ['phone triage', 'check in']) }
+  scope :not_assessments, -> { where.not(category: ['phone triage', 'check in']) }
```
timwis, line 46: lol @ name

app/models/need.rb
```diff
@@ -190,4 +196,13 @@ def self.sort_created_and_start_date(first, second)
     first.created_at <=> second.created_at
   end
   # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+
+  def valid_start_on?
```
timwis, line 200: FYI we could probably use the datetime validation gem Riccardo added. Of course this is fine too

app/views/triage/edit.html.erb
```diff
@@ -1,7 +1,8 @@
 
 <% if @contact_needs.errors.any? || @contact.errors.any? %>
```
timwis, line 2: I assume changes to this file is just formatting/linting/rubocop fixes?

config/application.rb
```diff
@@ -19,5 +19,6 @@ class Application < Rails::Application
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
+    config.action_view.field_error_proc = proc { |html_tag, _instance| html_tag.html_safe }
```
timwis, line 22: Just curious what this is?

app/controllers/assessments_controller.rb
jamiebuckley: Ah, yes.  Good point

app/models/need.rb
jamiebuckley: direct and to the point :smile:

app/views/triage/edit.html.erb
jamiebuckley: Correct, rubymine formatting after I had added and then removed something

app/controllers/assessments_controller.rb
timwis: We also want to associate the author of the note. At the moment it appears there's no category (even general) or author on the generated note.

timwis:
Hm, non-assessment needs don't seem to be showing on a person's profile for me. I have to hop on a call otherwise I'd diagnose further

config/application.rb
jamiebuckley: Rails default validation does something really annoying, where it wraps invalid fields in a
```
<div class="field_with_errors">
  <your original field/>
</div>
```
This tells it to just output the field as is when there is an error for the field

app/controllers/assessments_controller.rb
jamiebuckley: good point, forgot to do that

jamiebuckley:
Screenshot from 2020-04-29 15-32-11
Can't recreate the assessment vs non-assessment needs here, would be great if someone else could check this too

app/views/assessments/new.html.erb
```diff
@@ -0,0 +1,108 @@
...
+            </div>
+            <div class="radio">
+              <%= form.radio_button :category, 'check in', {:checked => @need[:category] == 'Check In'} %>
+              <%= form.label :category_check_in, "Check in", class: "radio__label" %>
+            </div>
+          </div>
+          <br/>
+
+          <div class="field">
+            <%= form.label :name, "Task body", class: "radio__label" %>
```
timwis, line 82: Is this the text and class you intended for this label?

app/views/assessments/new.html.erb
```diff
@@ -0,0 +1,108 @@
...
+            <%= form.text_field :name, placeholder: 'Describe the support action' %>
```
timwis, line 83: We probably only want this for scheduling, right? If you're logging one you've just completed, you instead want to leave a note about the call you just had.

app/views/assessments/new.html.erb
jamiebuckley: The class - no.  For the text, 'task body' is what it is set to in the locale - and therefore the error message.

app/controllers/assessments_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    @users = User.all
+    @roles = Role.all
+    render :new
+  end
+
+  def schedule_assessment
+    @need = Need.new(assessment_params.merge(contact_id: @contact.id))
+    @note = Note.new
```
timwis, line 47: I don't think you need this one, right?

jamiebuckley: it was actually simpler to include an @note rather than dealing with lots of conditional logic in the 'errors' section, e.g. `pluralize(@need.errors.count + @note.errors.count, "error")` 

The alternative would be to wrap the errors section in a `<% if type == 'log'>` section instead if you think that's better

app/views/assessments/new.html.erb
jamiebuckley: Fair point, have changed it to that

app/controllers/assessments_controller.rb
timwis: Oh! No worries I just assumed it was a leftover from copying & pasting

app/views/assessments/new.html.erb
```diff
@@ -0,0 +1,108 @@
...
+
+          <% if @type == 'log' %>
+            <%= fields_for :note, @note do |notes_fields| %>
+              <div class="field">
+                <%= notes_fields.text_area :body, placeholder: 'Call notes' %>
+              </div>
+            <% end %>
+          <% elsif @type == 'schedule' %>
+            <div class="field">
+              <%= form.label :name, "Task body", class: "field__label" %>
```
timwis, line 89: I think the label is meant to be "Description" or even "Describe the support action" like you have in the placeholder. We don't use the term "Task" in the app so this may be confusing otherwise.

app/views/assessments/new.html.erb
```diff
@@ -0,0 +1,108 @@
...
+
+              <% if @type == 'log' %>
+                <%= form.select :status, [%w(Complete complete)], {}, {class: "need-actions__select", readonly: true} %>
+              <% else %>
+                <%= form.select :status, need_statuses, {}, {class: "need-actions__select"} %>
+              <% end %>
+            </div>
+          </div>
+
+          <div class="field">
```
timwis, line 68: This should be wrapped in a `<fieldset>` and the label should be a `<legend>` (example in /views/contacts/_form_fields.erb.html) in order to use the app's radio styling. But feel free to leave it and create a card to fix that in the backlog, as it's just a minor styling issue.

app/views/assessments/new.html.erb
```diff
@@ -0,0 +1,108 @@
...
+            <span>Assessment Type</span>
+            <div class="radio">
+              <%= form.radio_button :category, 'phone triage', {:checked => @need[:category] == 'Phone triage'} %>
+              <%= form.label :category_phone_triage, "Phone Triage", class: "radio__label" %>
+            </div>
+            <div class="radio">
+              <%= form.radio_button :category, 'check in', {:checked => @need[:category] == 'Check In'} %>
```
timwis, line 75: Just pointing out this line uses a capital `I` and the next line uses a lowercase `i`. Not sure if that's an issue for this `==` or not.

jamiebuckley:
Thanks for the comments @timwis and the review - will fix the comments tomorrow in another PR

app/views/assessments/new.html.erb
jamiebuckley: No problem, agree - didn't seem the right terminology.  I'll track down what's using the locales message currently to see why we've set it to task body

PR #191, opened 2020-04-29
Combine the 'assign to user' and 'assign to team' dropdowns

timwis:
On the show need page, bulk assignment, and filters

This will have broken some tests, as expected, but thought I'd share my approach before diving into fixing tests.

### DIFF
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -2,8 +2,7 @@ class AssessmentsController < ApplicationController
   before_action :set_contact, only: %i[new create]
 
   def new
-    @users = User.all
-    @roles = Role.all
+    @assigned_to_options = construct_assigned_to_options
     @type = %w[log schedule].include?(type_param) ? type_param : 'log'
 
     @need = Need.new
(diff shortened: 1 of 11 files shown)


jamiebuckley:
Nice, this looks a good approach

app/controllers/assessments_controller.rb
```diff
@@ -61,10 +58,20 @@ def type_param
...
   def assessment_params
-    params.require(:need).permit(:user_id, :role_id, :name, :is_urgent, :status, :category, :status, :start_on)
+    params.require(:need).permit(:assigned_to, :name, :is_urgent, :status, :category, :status, :start_on)
   end
 
   def notes_params
     params.require(:note).permit(:body)
   end
+
+  def construct_assigned_to_options
```
eddieleeper-nb, line 68: This is duplicated in both controllers.
Also wondering if this is an option for caching (ok premature optimisation, but certainly an option for the future once someone has done some performance tests)

app/controllers/needs_controller.rb
```diff
@@ -74,6 +71,16 @@ def set_contact
   end
 
   def need_params
-    params.require(:need).permit(:name, :status, :user_id, :role_id, :category, :is_urgent, :lock_version)
+    params.require(:need).permit(:name, :status, :assigned_to, :category, :is_urgent, :lock_version)
+  end
+
+  def construct_assigned_to_options
```
eddieleeper-nb, line 77: as this is duplicate code I wonder if it's better placed in a view helper rather than the controller?

PR #192, opened 2020-04-29
Move export and filter buttons to the right, above the table

slav-petkov-saorsa:
Background:
https://trello.com/c/jUyj4aiB/436-move-export-and-filter-buttons-to-the-right-above-the-table

Solution:
- Moved export and filter buttons to  the right, above the temple.
- Since text-overflow doesn't seem to work correctly on select element, except for Firefox, I made the roles assignment boxes wider, so not to clip the displayed text.

image

### DIFF
--- a/app/assets/stylesheets/_globals.scss
+++ b/app/assets/stylesheets/_globals.scss
@@ -45,6 +45,29 @@ select::-ms-expand {
   display: none;
 }
 
+.dropdown {
+  margin-left: 5px;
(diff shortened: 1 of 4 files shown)


slav-petkov-saorsa:
[Move export and filter buttons to the right, above the table](https://trello.com/c/jUyj4aiB/436-move-export-and-filter-buttons-to-the-right-above-the-table)

PR #195, opened 2020-04-29
420 team support links

eddieleeper-nb:


app/helpers/application_helper.rb
```diff
@@ -26,4 +26,8 @@ def friendly_boolean(value)
       '—'
     end
   end
+
+  def current_role_id
+    current_user.role&.id
```
jamiebuckley, line 31: Thinking about it - I think this could be:
```
current_user.assign_role_if_empty (Which I suppose I should add a ! to)
current_user.role.id
```
Assign role if empty will provide a noisy error if no valid role can be assigned to the user - which I think is decent behaviour for now - whereas there is the potential for the user here not to see the values they expect on first log in

jamiebuckley: Just tried it out - this isn't urgent - there are already policies that execute and set this role before this code would be executed

PR #197, opened 2020-04-30
Feature/completed needs collapsed

go8soft:
Task: https://trello.com/c/D6dObIiI/441-show-completed-needs-like-we-show-completed-assessments-collapsed


app/javascript/packs/completed-needs.js
```diff
@@ -0,0 +1,10 @@
+document.querySelectorAll('.completed-need').forEach(e => e.style.display = 'none');
```
timwis, line 1: Didn't Jamie add collapsing code for assessments already? Perhaps it would make sense to use that since it's the same functionality?

go8soft: Good point, thanks. Not sure will have time to collect the logic from two spaces in one in next 30 mins.

app/javascript/packs/assessments.js
```diff
@@ -3,8 +3,15 @@ Array.prototype.slice.call(document.querySelectorAll('.need--status_complete')).
 let showCompletedNeeds = false;
 const toggleDisplay = (show) => show ? 'table-row' : 'none';
 
-document.getElementById('toggle-visibility-completed-assessments').addEventListener('click', (e) => {
+const toggleVisibilityCompletedAssessments = document.getElementById('toggle-visibility-completed-assessments')
+toggleVisibilityCompletedAssessments.addEventListener('click', (e) => {
     showCompletedNeeds = !showCompletedNeeds;
+    const table = toggleVisibilityCompletedAssessments.parentNode.parentNode
```
rik86future, line 9: I think it's better to explicitly add semicolons going forward

go8soft: We have a lot js without `;` also `let` and `const` for example. We expect browser to support that.

rik86future: no I know the browser supports it, it's more for other issues can come out of it sometimes. I read an article about it some time ago, basically if you let js put the semicolons for you sometimes they get put in the wrong place, so it's better to add them explicitly

go8soft: You are right. It is good to put `;` on places you mention. For example around `return`. For all other places it should be project based.

PR #198, opened 2020-04-30
Small styling and layout fixes for assessments

jamiebuckley:


app/views/assessments/new.html.erb
```diff
@@ -24,10 +24,10 @@
 
             <ul>
               <% @need.errors.full_messages.each do |message| %>
-                <li><%= message %></li>
+                <li>Assessment <%= message.downcase %></li>
```
jamiebuckley, line 27: It was confusing on the errors which field it was referring to, e.g. it would just say "Body is required" rather than "Note body is required"

app/views/assessments/new.html.erb
```diff
@@ -65,17 +65,17 @@
             </div>
           </div>
 
-          <div class="field">
-            <span>Assessment Type</span>
+          <fieldset class="field-section">
+            <legend class="field-section__title">Assessment Category</legend>
```
jamiebuckley, line 69: The figma designs say "Assessment Type", but the field is 'category' and we are referring to it as category elsewhere in the app - makes sense to me to keep category here for the time being

app/views/assessments/new.html.erb
```diff
@@ -65,17 +65,17 @@
...
             <div class="radio">
               <%= form.radio_button :category, 'phone triage', {:checked => @need[:category] == 'Phone triage'} %>
               <%= form.label :category_phone_triage, "Phone Triage", class: "radio__label" %>
             </div>
             <div class="radio">
-              <%= form.radio_button :category, 'check in', {:checked => @need[:category] == 'Check In'} %>
+              <%= form.radio_button :category, 'check in', {:checked => @need[:category] == 'Check in'} %>
```
jamiebuckley, line 75: Small issue that would prevent your choice from being kept when the form was re-rendered after validation errors

timwis: ah! that happened to me but I couldn't reproduce so I assumed I was going mad

PR #200, opened 2020-04-30
remove Check-in and Phone triage needs from triage page

rik86future:


rik86future:
[(misc) Don't let users add Phone triage or Check-in need types on Triage form](https://trello.com/c/vz91etkd/407-misc-dont-let-users-add-phone-triage-or-check-in-need-types-on-triage-form)

app/models/need.rb
```diff
@@ -173,7 +175,7 @@ def self.dynamic_fields
   end
 
   def self.categories_for_triage
-    categories.except('Other')
+    categories.except('Other').reject { |_k, v| v.in? ASSESSMENT_CATEGORIES }
```
timwis, line 178: Little trick: you could do `categories.except('Other', *ASSESSMENT_CATEGORIES)` (it will expand the array to be arguments)

features/step_definitions/support_action_steps.rb
```diff
@@ -4,6 +4,15 @@
   choose_yes_on_support_action(page, support_action)
 end
 
+When('I add support actions') do
+  visit "/contacts/#{@contact.id}"
+  click_link 'Add Needs'
```
timwis, line 9: Oh shoot, did we leave the word 'Needs' on this list? We should probably change it to 'Add support actions' shouldn't we 😬

PR #204, opened 2020-04-30
Add rake tasks to migrate needs assigned to dummy users to corresponding role

jamiebuckley:


timwis:
We're going to do this manually in the UI

PR #208, opened 2020-04-30
Fix user roles scripts, preventing assigining duplicate roles to a user

jamiebuckley:


lib/tasks/users_roles.rake
```diff
@@ -29,8 +29,9 @@ namespace :upgrade do
       admin_role = Role.find_by!(role: 'manager')
       agent_role = Role.find_by!(role: 'agent')
       ActiveRecord::Base.transaction do
-        User.all.find_each do |user|
-          user.roles << (user.admin? ? admin_role : agent_role)
+        User.all.includes(:roles).find_each do |user|
+          expected_role = user.admin? ? admin_role : agent_role
+          user.roles << expected_role unless user.roles.include?(expected_role)
```
go8soft, line 34: Dont make sense for me  -`user.roles = user.roles.uniq` should remove duplicated roles?

jamiebuckley: Ah, it fails on addition, rather than saving

go8soft: You are right. After the validations it is impossible to add the same role twice to the user. The result will be rails validation error. With that fix we will add only unique role to the user.

PR #214, opened 2020-05-01
467 nhs number should be nhs number

rik86future:
Fixes 3 simple UI issues:
- nhs number should be NHS number
- Show consents on profile 
- Assessment Errors validations

image

### DIFF
--- a/app/models/need.rb
+++ b/app/models/need.rb
@@ -35,9 +35,9 @@ class Need < ApplicationRecord
 
   validates :category, presence: true
   validates :name, presence: true
-  validate :valid_start_on?
+  validate :start_required_for_assessment
 
-  def valid_start_on?
+  def start_required_for_assessment
     return unless start_on.nil? && ASSESSMENT_CATEGORIES.include?(category.to_s.downcase) ||
                   start_on.nil? && category.nil?
 
--- a/app/models/need.rb
+++ b/app/models/need.rb
@@ -240,6 +240,7 @@ class Need < ApplicationRecord
 
   # This sort method is to first sort future needs (where start_on > today) to the bottom of the list
   # and then sort by created_at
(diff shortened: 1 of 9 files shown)


spec/factories/needs.rb
```diff
@@ -7,7 +7,7 @@
     contact
     category { Need.categories.values.sample }
     name { Faker::Lorem.sentence }
-    start_on { [nil, Faker::Date.between(from: 2.days.ago, to: 6.days.from_now)].sample }
+    start_on { [nil, Faker::Date.between(from: 1.days.from_now, to: 6.days.from_now)].sample }
```
timwis, line 10: This had a `from` date in the past so that we could see needs that had a start date that had started. Any reason we need to change it?

rik86future: oh yeah I will revert it, no need to change

app/models/need.rb
```diff
@@ -32,11 +34,16 @@ class Need < ApplicationRecord
...
 
   validates :category, presence: true
+  validates :name, presence: true
+  validate :valid_start_on?
 
-  ASSESSMENT_CATEGORIES = ['phone triage', 'check in'].freeze
-
-  # validates :food_priority, inclusion: { in: %w[1 2 3] }, allow_blank: true
-  # validates :food_service_type, inclusion: { in: ['Hot meal', 'Heat up', 'Grocery delivery'] }, allow_blank: true
+  # rubocop:disable Style/GuardClause, Style/IfUnlessModifier
```
eddieleeper-nb, line 40: maybe this works - then no need to disable the cop.
```
  validate :start_required_for_assessment

  def start_required_for_assessment
    return unless start_on.nil? && ASSESSMENT_CATEGORIES.include?(category.to_s.downcase) ||
                  start_on.nil? && category.nil?

    errors.add(:start_on, 'must be provided')
  end
```

rik86future: thanks, done

PR #216, opened 2020-05-01
Add needs default status

go8soft:


timwis:
Do we need this now that we have it in a database constraint?

go8soft:
> Do we need this now that we have it in a database constraint?

The change will allow AR to know what we have in the DB before trying to work with it.
For example if you build and object or use `rails c` it will be very helpful. Even more - usually we have the logic in the model and some of the logic in DB.

go8soft:
> Approved; consider adding a test.

Thanks @eddieleeper-nb I created the default status build object test.

PR #219, opened 2020-05-01
UI fixes

rik86future:


app/controllers/assessments_controller.rb
```diff
@@ -29,7 +29,7 @@ def set_contact
   end
 
   def log_assessment
-    @need = Need.new(assessment_params.merge(contact_id: @contact.id, name: 'Logged assessment'))
+    @need = Need.new(assessment_params.merge(contact_id: @contact.id, name: 'Logged assessment', start_on: Date.today))
```
timwis, line 32: Why are we adding a start date? It's not a required field

app/models/need.rb
```diff
@@ -32,11 +34,15 @@ class Need < ApplicationRecord
                    'Other': 'other' }
 
   validates :category, presence: true
+  validates :name, presence: true
+  validate :start_required_for_assessment
 
-  ASSESSMENT_CATEGORIES = ['phone triage', 'check in'].freeze
+  def start_required_for_assessment
```
timwis, line 40: Why is this required?

spec/factories/needs.rb
```diff
@@ -7,7 +7,9 @@
     contact
     category { Need.categories.values.sample }
     name { Faker::Lorem.sentence }
-    start_on { [nil, Faker::Date.between(from: 2.days.ago, to: 6.days.from_now)].sample }
+    start_on do
+      Faker::Date.between(from: 2.days.ago, to: 6.days.from_now) if category.in?(['check in', 'phone triage'])
```
timwis, line 11: Since start_on isn't a required field, shouldn't we keep some null ones in?

app/views/shared/_profile-accordion-edit.html.erb
```diff
@@ -112,7 +115,35 @@
...
+  </div>
+
+  <button class="accordion__control" aria-expanded="true" id="contact-additional-info-accordion">
+    <h2 class="accordion__title">Consents</h2>
+  </button>
+  <div class="accordion__content">
+    <dl class="details-list">
+      <dt class="details-list__caption">Do not call</dt>
+      <dd class="details-list__value" id="no_calls_flag">
+        <%= friendly_boolean @contact.no_calls_flag %>
```
jamiebuckley, line 130: Would we also want these to be editable here?

timwis:
Would you mind separating out the UI tweaks from the data validation stuff? We know we want the UI tweaks straight away, but the data validation needs more discussion I think

app/views/shared/_profile-accordion-edit.html.erb
timwis: Yes

timwis: From Laura just now:
> when you click on add support action and it opens up the needs and the contact details to edit you should be ablet to add consent there
 but it currently doesn't appear
you can only add it by clicking edit their details
this isn't going to solve the current problem because users won't see it on calls

rik86future:
@timwis and @jamiebuckley 
Regarding this one, I only cherry picked what gone into master and already approved. If I do changes now not only I will create merge conflicts but also the logic will work differently in master and release 2, perhaps we should not include it

here is the original PR: https://github.com/wearefuturegov/beacon/pull/214

If you don't feel like merging it I can close it, but I am not editing it

PR #224, opened 2020-05-01
Assignment usability

slav-petkov-saorsa:
Background:
https://trello.com/c/zSdkJ9Go/447-assignment-usability

Solution:
- When the assign to select box is rendered, it has initially the disabled attribute applied to it.
- When the user manipulates a need checkbox, the select box will be enabled/disabled based on the count of checked needs.
- When a user un-checks all check boxes, the Assign All check box get un-checked as well.


slav-petkov-saorsa:
[Assignment usability](https://trello.com/c/zSdkJ9Go/447-assignment-usability)

PR #228, opened 2020-05-04
BugFix: Food manager receives error

jamiebuckley:
An explanation of the issue:
There are three 'mutations' of the active record query that get performed.
1. The policy scope, which limits the records a user can see to those assigned to the right teams/users
2. The aggregations, which add extra calculated fields from subqueries
3. The filter/sort

Given that there is a fairly complex interplay between these queries, the simplest solution is to take the policy queries, and move them into a `where in (subquery)` query.  This means that the sorting issues - caused by a mismatch of distinct/aggregate columns, should be fixed, and the overall SQL structure is easier to understand.

### DIFF
--- a/app/models/need.rb
+++ b/app/models/need.rb
@@ -89,6 +89,10 @@ class Need < ApplicationRecord
     end
   }
 
+  scope :order_by_created_at, lambda { |direction|
+    order("needs.created_at #{direction}")
+  }
+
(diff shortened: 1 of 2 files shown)


rik86future:
[Food Manager recieves error when sorting Support Actions table by date](https://trello.com/c/iVUsAI5R/92-food-manager-recieves-error-when-sorting-support-actions-table-by-date)

PR #233, opened 2020-05-04
54 add delete needs

rik86future:


rik86future:
["Delete" and "Edit" Support actions, Call Logs and Notes created in error- Contact Agents and Managers](https://trello.com/c/DkPSRrjC/54-delete-and-edit-support-actions-call-logs-and-notes-created-in-error-contact-agents-and-managers)

app/models/need.rb
```diff
@@ -80,6 +83,8 @@ def start_required_for_assessment
 
   scope :filter_by_status, ->(status) { where(status: status) }
 
+  scope :created_by, ->(user_id) { joins(:versions).where('whodunnit = ? and event = ?', user_id.to_s, 'create') }
```
jamiebuckley, line 86: Wonder if we should create an index for this

app/policies/need_policy.rb
```diff
@@ -46,6 +46,20 @@ def show?
     Pundit.policy_scope!(@user, Need).where(id: @record.id).exists?
   end
 
+  def destroy?
+    # managers can always destroy
+    return true if admin?
+
+    # check ownership of need record
+    need = Pundit.policy_scope!(@user, Need).where(id: @record.id)
```
jamiebuckley, line 54: Policy scope for certain roles (MDT, Food Hub, Council Services), restricts viewable needs to those that are assigned to them, their team, or their role.

This filtering is done by the policy scope - so `Pundit.policy_scope!(@user, Need).where(id: @record.id).exists?` will return false if it is not assigned to one of those three that put it in the policy scope, even if the user created it.

rik86future: ok I get it now, thanks

rik86future: @jamiebuckley  amended now

jamiebuckley:
Looks great

PR #235, opened 2020-05-05
Feature/cleanup production data food priority 3

go8soft:
Task: https://trello.com/c/Xl2iFeiP/94-script-to-cleanup-production-data-food-priority-3
Please check that there is a backup copy of prod db before run the task.

To run the task: 
```
i-have-i-need$ bundle exec rake cleanup_needs_by_priority:food_priority_3
```

### DIFF
--- a/lib/tasks/cleanup_needs_by_priority.rake
+++ b/lib/tasks/cleanup_needs_by_priority.rake
@@ -1,0 +1,6 @@
+namespace :cleanup_needs_by_priority do
+  desc "Detele needs and notes records for needs.supplemental_data.food_priority == 3."
+  task food_priority_3: :environment do
+    p Need.where('supplemental_data @> ?', {food_priority: "3"}.to_json).destroy_all
+  end
+end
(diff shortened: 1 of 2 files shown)


lib/tasks/cleanup_needs_by_priority.rake
```diff
@@ -0,0 +1,6 @@
+namespace :cleanup_needs_by_priority do
+  desc 'Detele needs and notes records for needs.supplemental_data.food_priority == 3.'
+  task food_priority_3: :environment do
+    p Need.where('supplemental_data @> ?', { food_priority: '3' }.to_json).destroy_all
```
jamiebuckley, line 4: Will this also work if the need has notes?

jamiebuckley: yep tested locally, seems to work there

go8soft: yes, `destroy_all` takes care of the related data

PR #237, opened 2020-05-05
Change category need ad s bulk action

go8soft:
Task: https://trello.com/c/tW6HfDQp/22-quick-update-the-status-of-multiple-needs-use-case-food


rik86future:
[Quick update the status of multiple needs (use case: food)](https://trello.com/c/tW6HfDQp/22-quick-update-the-status-of-multiple-needs-use-case-food)

PR #239, opened 2020-05-07
93 consent checkbox bug not appearing on profile after user has entered data into it

rik86future:
Fix several checkboxes in the triage form
Fix styling in triage form


rik86future:
[Consent Checkbox bug- not appearing on profile after user has entered data into it](https://trello.com/c/2eFsmbeP/93-consent-checkbox-bug-not-appearing-on-profile-after-user-has-entered-data-into-it)

PR #241, opened 2020-05-11
80 system admin restore deleted records

rik86future:


app/controllers/needs_controller.rb
```diff
@@ -89,6 +110,31 @@ def request_destroy_note
...
   end
 
+  def restore_need
+    need = policy_scope(Need).deleted.where(id: params[:id])
+    if need.exists?
+      need_name = need.first.category
+      Rails.logger.info("Restored need '#{params[:id]}'")
+      need.first.restore(:recursive => true)
+      redirect_to deleted_needs_path(order: "deleted_at", order_dir: "DESC"), 
+        notice: "Restored '#{need_name}' see <a href='/needs/#{need.first.id}'>here</a>"
```
jamiebuckley, line 120: could use the URL helpers here, e.g. `#{need_path(need.first.id)}`

jamiebuckley: And potentially:
`view_context.link_to` helper as well

app/views/layouts/application.html.erb
```diff
@@ -33,7 +33,7 @@
 
       <!-- Notifications -->
       <% flash.each do |type, msg| %>
-        <div class="<%= type %>"><%= msg %></div>
+        <div class="<%= type %>"><%= msg.html_safe %></div>
```
jamiebuckley, line 36: Perhaps I'm being paranoid, but wonder if we should run this through something like https://github.com/rgrove/sanitize first

app/controllers/needs_controller.rb
rik86future: done

app/views/layouts/application.html.erb
rik86future: fair, I used the built in SanitizeHelper instead

rik86future:
[System admin restore "deleted" records](https://trello.com/c/3M3eMEnC/80-system-admin-restore-deleted-records)

PR #242, opened 2020-05-11
Feature/103 table to show needs ive created but left unassigned

jamiebuckley:
Currently this is implemented as another dashboard link in the sidebar
Screenshot from 2020-05-12 09-50-42
There are designs that display this embedded in a page, but as it is part of a wider dashboard redesign, that seems better left until those are finalized

### DIFF
--- a/app/controllers/needs_controller.rb
+++ b/app/controllers/needs_controller.rb
@@ -18,7 +18,7 @@ class NeedsController < ApplicationController
              else
                @assigned_to_options = construct_assigned_to_options
                policy_scope(Need).started
-    end
+             end
 
     @needs = @needs.filter_and_sort(@params.slice(:category, :assigned_to, :status, :is_urgent), @params.slice(:order, :order_dir))
     @needs = @needs.page(params[:page]) unless request.format == 'csv'
--- a/app/controllers/needs_controller.rb
+++ b/app/controllers/needs_controller.rb
@@ -8,14 +8,20 @@ class NeedsController < ApplicationController
   helper_method :get_param
 
   def index
-    @params = params.permit(:assigned_to, :status, :category, :page, :order_dir, :order, :commit, :is_urgent)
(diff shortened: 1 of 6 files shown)


jamiebuckley:
[Table to show needs I've created but left 'unassigned'](https://trello.com/c/1lkaPsQY/103-table-to-show-needs-ive-created-but-left-unassigned)

features/list_support_actions.feature
```diff
@@ -46,4 +46,20 @@ Feature: List support actions
...
+  Scenario: See unassigned needs I have created in the created and unassigned list
+    Given a unique resident
+    And I add support actions "Dog walking"
+    And I submit the add support actions form
+    When I go the the created and unassigned support action list
+    Then I see the support action for category "Dog walking" in the results
+
+  @javascript
+  Scenario: Do not see assigned needs I have created in the created and unassigned list
+    Given a unique resident
```
rik86future, line 60: You could compress this into "Given I have created a support action "Dog Walking""
Which calls
```
step  a unique resident
step I add support actions "Dog walking"
step I submit the add support actions form
step I assign the support action to another user
```

It does not make any difference expect for it reading a bit easier

app/controllers/needs_controller.rb
```diff
@@ -8,14 +8,20 @@ class NeedsController < ApplicationController
   helper_method :get_param
 
   def index
-    @params = params.permit(:assigned_to, :status, :category, :page, :order_dir, :order, :commit, :is_urgent)
+    @params = params.permit(:assigned_to, :status, :category, :page, :order_dir, :order, :commit, :is_urgent, :created_by_me)
     @users = User.all
```
rik86future, line 12: I am not sure if we still need there @users and @roles variables here

jamiebuckley: Good spot - these have been superceded by the 'construct_assigned_to_options' method, will remove them

features/list_support_actions.feature
jamiebuckley: Fair point, have updated

PR #247, opened 2020-05-12
14 deactivate users

rik86future:
image

image

### DIFF
--- a/app/assets/stylesheets/table.scss
+++ b/app/assets/stylesheets/table.scss
@@ -32,6 +32,16 @@
                     }
                 }
             }
+            &.deleted-row{
+                background: transparentize($grey, 0.9);
+                &:hover{
+                    background: transparentize($grey, 0.8) !important;
+                }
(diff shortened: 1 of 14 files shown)


rik86future:
[Deactivate users](https://trello.com/c/W9oatnb5/14-deactivate-users)

jamiebuckley:
The core problem with deleting users is that they can be the owner of needs, notes etc.  Currently when deleting a user, needs that are assigned to them seem to default to 'unassigned' in the index view, presumably because we can't read a deleted user's name.

Also, on the needs detail view, the assigned to box displays 'Assigned to me' as the deleted users aren't appearing in that assigned list, so the need displays as 'Assigned to me'.

jamiebuckley:
Maybe this isn't critical though, I'd think a process for deactivating a user could be:
1. Deactivate the user
2. Search for actions assigned to that user (which seems to work fine, although it displays 'Unassigned')
3. Re-assign those needs

PR #251, opened 2020-05-13
Feature/bulk assignment notify emails

jamiebuckley:
Screenshot from 2020-05-13 14-40-56


jamiebuckley:
Wording will need to be improved to be singular/plural, but the principle works

PR #253, opened 2020-05-13
Contact to role

go8soft:
Task: https://trello.com/c/4hhK8ugN/108-allow-allocating-a-contact-to-a-role

### DIFF
--- a/app/controllers/needs_controller.rb
+++ b/app/controllers/needs_controller.rb
@@ -192,13 +192,4 @@ class NeedsController < ApplicationController
     assigned_to = "user-#{current_user.id}" if assigned_to == 'assigned-to-me'
     assigned_to
   end
-
-  def construct_assigned_to_options
-    users = User.all.order(:first_name, :last_name)
-
-    {
-      'Teams' => Role.construct_assigned_to_options,
-      'Users' => users.map { |user| [user.name_or_email, "user-#{user.id}"] }
-    }
-  end
 end
--- a/app/controllers/needs_controller.rb
+++ b/app/controllers/needs_controller.rb
@@ -19,6 +19,7 @@ class NeedsController < ApplicationController
 
     @needs = @needs.filter_and_sort(@params.slice(:category, :assigned_to, :status, :is_urgent), @params.slice(:order, :order_dir))
     @needs = @needs.page(params[:page]) unless request.format == 'csv'
+    @assigned_to_options_with_deleted = construct_assigned_to_options(true)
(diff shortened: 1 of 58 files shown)


app/controllers/needs_controller.rb
```diff
@@ -192,4 +192,13 @@ def assigned_to_me(assigned_to)
     assigned_to = "user-#{current_user.id}" if assigned_to == 'assigned-to-me'
     assigned_to
   end
+
+  def construct_assigned_to_options
```
rik86future, line 196: I think this is a merge issue, there is already another `construct_assigned_to_options` in this controller, outside of the private methods

go8soft: Sure, working on it.

slav-petkov-saorsa: When I try to start the app. Maybe overlaps with Riccardo's comment

image

app/javascript/packs/load-service-panel.js
```diff
@@ -0,0 +1,35 @@
+const updateLeadService = (contact_id, lead_service_id, lead_service_note) => {
+  $.ajax({
```
rik86future, line 2: as I was saying before I don't think we need all this javascript, we can try to use the built in `form remote:true` options rails 5 has, and just re-render the partial, do you think you can try that way?

app/controllers/needs_controller.rb
go8soft: Yes, I tried to resolve the conflicts using Git web and did mistake. Working to fix it.

slav-petkov-saorsa: Select Box is to small, And didn't we have an avatar in the designs?

image

go8soft: This is not what I did with the styling. Strange where the styles lost.

slav-petkov-saorsa: And also, the buttons are on the right hand side.

And we also have a checkbox for sending emails. Maybe we decided to leave it out?

go8soft: About the avatar I dont set avatar, because we dont have avatar for different roles and it will always be the same.

go8soft: > And also, the buttons are on the right hand side.
> 
> And we also have a checkbox for sending emails. Maybe we decided to leave it out?

I think email sending is part of something that @jamiebuckley already did. Not sure how to connect it here.

PR #264, opened 2020-05-18
Add edit note

go8soft:
Task: https://trello.com/c/Bi74hQoQ/106-edit-needs-and-notes-after-they-have-been-created-placeholder


app/controllers/notes_controller.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class NotesController < ApplicationController
+  before_action :load_note, only: %i[show edit update]
```
jamiebuckley, line 4: Minor thing for consistency, we have usually called this type of method `set_x`, e.g. `set_contact`, `set_need`

go8soft: Thanks!

PR #265, opened 2020-05-18
add assessment page

rik86future:


app/controllers/assessments_controller.rb
```diff
@@ -24,8 +36,35 @@ def create
 
   private
 
+  def contact_needs(need_id)
+    needs = Need.where(assessment_id: need_id)
+    create_contact_needs if needs.empty?
```
rik86future, line 41: to be completed

app/controllers/assessments_controller.rb
```diff
@@ -24,8 +36,35 @@ def create
...
+  end
+
+  def create_contact_needs
```
rik86future, line 44: will refactor

PR #269, opened 2020-05-20
Run cucumber features in parallel to reduce build time

jamiebuckley:


rik86future:
Parallel tests can possibly cause data conflicts be mindful of that, also the concurrent features should not run in parallel

jamiebuckley:
Yep, this was an experiment during downtime during builds etc. yesterday, I'll probably close this, primarily as we are unlikely to have to time, but also it doesn't give as much of a time saving on github as it does locally.

For data issues, it creates separate databases for each test which is quite clever, but also the source of a few pain points.

PR #278, opened 2020-05-21
Add delete items single nav bar

go8soft:
Task: https://trello.com/c/fuQMmlmX/139-change-delete-needs-notes


slav-petkov-saorsa:
- Drop down doesn't match the rest of the drop downs look and feel.
image

- When i hover over a row in the deleted items table, I can see the cursor changing to a pointer. And I expect a navigation to occur when I click on the row. But nothing is happening. Should you be able to view the details of a deleted need? If no, there shouldn't be a pointer appearing when I hover over the row.

go8soft:
>     * Drop down doesn't match the rest of the drop downs look and feel.
>       image
> 
>     * When i hover over a row in the deleted items table, I can see the cursor changing to a pointer. And I expect a navigation to occur when I click on the row. But nothing is happening. Should you be able to view the details of a deleted need? If no, there shouldn't be a pointer appearing when I hover over the row.

Thanks for the notes.

- Yes, we can try to make it as rest of the drop downs.

-  It is not related with this PR changes.

go8soft:
>     * Drop down doesn't match the rest of the drop downs look and feel.
>       image
> 
>     * When i hover over a row in the deleted items table, I can see the cursor changing to a pointer. And I expect a navigation to occur when I click on the row. But nothing is happening. Should you be able to view the details of a deleted need? If no, there shouldn't be a pointer appearing when I hover over the row.

- Did changes regarding the point.

PR #297, opened 2020-05-22
Needs future navigation

go8soft:
Task: https://trello.com/c/jhHRiwsS/155-needs-nav-future-triages-and-check-ins


jamiebuckley:
The start on filter can't be cleared currently, can you add a 'any time' with value null etc. option to the select?

PR #300, opened 2020-05-26
Feature/163 dashboard changes manager view

jamiebuckley:


app/models/need.rb
```diff
@@ -57,21 +57,25 @@ def start_required_for_assessment
...
-        date = nil
-      end
+    next all if start_on.nil? || start_on == 'any_time'
+    case start_on
+    when 'today'
+      where('start_on >= ? and start_on <= ?', Date.today, Date.today.end_of_day)
+    when 'tomorrow'
+      where('start_on >= ? and start_on <= ?', Date.today + 1.days, Date.today.end_of_day + 1.days)
+    when 'next_week'
+      where('start_on >= ? and start_on <= ?', Date.today + 1.days, Date.today.end_of_day + 7.days)
```
rik86future, line 67: I think you can just write `Date.tomorrow` but it does not matter

jamiebuckley: Ah good spot, yeah that would be better

PR #314, opened 2020-05-27
Allow user to choose if assignment email is sent using checkbox

rik86future:
image


app/services/needs_assignee_notifier.rb
```diff
@@ -1,8 +1,9 @@
 class NeedsAssigneeNotifier
   def self.notify_new_assignee(need)
-    return unless need.saved_change_to_user_id? || need.saved_change_to_role_id?
+    return unless need.saved_change_to_user_id? || need.saved_change_to_role_id? || need.send_email
```
jamiebuckley, line 3: I think this should maybe be `(x || y) && z` otherwise this will send an email if the send_email flag is true, even if one of the changes hasn't happened.

rik86future: I changed it

app/views/needs/_need-actions.html.erb
```diff
@@ -1,26 +1,33 @@
...
-    </div>
-<% end %>
+      <div class="need-actions__field" id="status-actions__field">
+      <%= f.label :status, 'Status', class: 'need-actions__label dropdown-label' %>
+      <%= f.select :status, need_statuses, {}, { class: "need-actions__select dropdown", "data-minimum-results-for-search" => "Infinity" } %>
+      </div>   
+    </div> 
+
+    <div class="send-email__field" id="send-email__field">
+      <%= f.check_box :send_email, {class: 'bulk-action-checkbox', checked: false,  id: "send-email"}, 'true', 'false' %>
```
jamiebuckley, line 28: should checked default to false here?

rik86future: yes it defaults to false, I think they want it to be false by default and optionally set it to "True" so they can send the email, otherwise users would have to un-tick the box every time...anyway not the best user experience

jamiebuckley: I'm just thinking that if all the checkboxes default to false, there is no point persisting it on the model

rik86future: I need to persist it because during the assessment we save that preference but the email is not sent until assessment completion...

jamiebuckley: I see, yeah that makes sense

app/views/assessments/assign.html.erb
```diff
@@ -69,6 +71,10 @@
                     <td>
                       <%= need_fields.select :status, need_statuses, {}, { class: "dropdown", "data-minimum-results-for-search" => "Infinity" } %>
                     </td>
+                    <td>
+                      <%= need_fields.check_box :send_email, {class: 'bulk-action-checkbox', checked: false,  id: "send-email"}, 'true', 'false' %>
```
jamiebuckley, line 75: Given the other conversation then - should the checked value be read here, to keep previous selections?

rik86: yes in this case

PR #320, opened 2020-05-28
Fix Assigned to me in Assigned dropdown in Assign section on assessme…

go8soft:
…nts flow

### DIFF
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -1,4 +1,4 @@
-class AssessmentsController < NeedsTableController
+class AssessmentsController < ApplicationController
   before_action :set_contact, only: %i[new create]
   before_action :set_assessment, only: %i[fail update_failure edit update assign update_assignment complete update_completion start]
   include AssigningConcern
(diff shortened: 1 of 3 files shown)


app/controllers/assessments_controller.rb
```diff
@@ -1,4 +1,4 @@
-class AssessmentsController < ApplicationController
+class AssessmentsController < NeedsTableController
```
rik86future, line 1: no, it inherits too much stuff from NeedsTableController, what do you need to inherit exactly?

go8soft: I need `assigned_to_me` from `NeedsTableController`

rik86future: in this case it's better to copy it as is

rik86future: otherwise you can add it to the AssigningConcern as a helper method and include AssigningConcern in the AssessmentsController

go8soft: Ready to go now? https://github.com/wearefuturegov/beacon/pull/320/commits/c1efb359c05febc90c997562d61cd43782f9bb00

PR #321, opened 2020-05-28
Feature/45 data validation

jamiebuckley:


app/models/contact.rb
```diff
@@ -12,7 +12,14 @@ class Contact < ApplicationRecord
   has_paper_trail
 
   validates :first_name, presence: true
```
rik86future, line 14: I think these would be fairly easy to test in the contact_spec.rb

jamiebuckley: yeah good point, will add tests for those

app/models/contact.rb
```diff
@@ -12,7 +12,14 @@ class Contact < ApplicationRecord
...
+  validates :surname, presence: true
   validates_date :date_of_birth, allow_nil: true, allow_blank: true
+  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }, if: :email?
+  validates :telephone, numericality: true, length: { minimum: 9, maximum: 11 }, if: :telephone?
```
rik86future, line 18: the drawback is that we don't allow spaces any more for telephones

app/models/contact.rb
```diff
@@ -12,7 +12,14 @@ class Contact < ApplicationRecord
   has_paper_trail
 
   validates :first_name, presence: true
+  validates :surname, presence: true
   validates_date :date_of_birth, allow_nil: true, allow_blank: true
+  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }, if: :email?
```
rik86future, line 17: a*@gmail.com is not a valid email but this regex seems to accept it

app/models/contact.rb
jamiebuckley: yes, that does introduce some complexities.  It's a little tricky validating length and also allowing spaces

app/models/contact.rb
jamiebuckley: This regex comes from:
https://html.spec.whatwg.org/multipage/input.html#e-mail-state-(type=email)

* is a valid part of an email
https://en.wikipedia.org/wiki/Email_address#Local-part

rik86future: ok

rik86future:
I think this PR is ok with the tests but I would be more comfortable to merge it if we also had the rake tasks to normalise existing data for prod...what do you think?

jamiebuckley:
I agree, if you are happy with the PR then we'll just leave it open, and I'll start work on the data identification scripts.

PR #331, opened 2020-06-02
Contact info reviewed

rik86future:
- Change contact search page to show:
Full Name, Address, Postcode, Date of Birth, Lead Service

- Change Contact details to hide: NHS Number info

- Remove export functionality


app/controllers/contacts_controller.rb
```diff
@@ -68,6 +70,16 @@ def update
     invalid_update
   end
 
+  def name_for_lead_service(role_id)
```
jamiebuckley, line 73: Does the role_id need to be passed in?
Could just be:
```
  Roles.find(lead_service_id).name if lead_service_id
```

jamiebuckley: or `lead_service_id.present? ? Roles.find(lead_service_id).name : 'No lead service'` etc.

rik86future: I am not sure how rails handles that but I was hoping to have a cached list of roles instead of querying it for every row, I thought it may be a bit faster

rik86future: so basically instead of querying

findById 10 times, it only query findAll roles 1 time and filter that list, I am not sure if the performance is noticeable

PR #344, opened 2020-11-20
Add test and trace account ID to contacts

userman123:


go8soft (changes requested):
Please add seed data for it.

PR #350, opened 2020-11-24
Add user restore

go8soft:


app/controllers/users_controller.rb
```diff
@@ -60,6 +60,7 @@ def update
   # DELETE /users/1
   def destroy
     user = User.find(params[:id])
+    authorize user
```
userman123, line 63: Curious as to why this was added, was it just missing from before or is it something to do with the new restore feature?

PR #355, opened 2020-11-30
Rework existing export

go8soft:
I reworked the existing import. Please take a look at how we show duplicated records error. I can also add new column in needs section(s) to show the name of the import to allow bulk actions on that needs.

You should also run `bundle install`, because I added new gem for xlsx.

### DIFF
--- a/app/models/imported_item.rb
+++ b/app/models/imported_item.rb
@@ -3,7 +3,7 @@
 class ImportedItem < ApplicationRecord
   include Filterable
 
-  require 'csv'
+  require 'roo'
 
   validates_uniqueness_of :name
   validates :name, presence: true, allow_blank: false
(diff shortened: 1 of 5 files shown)


app/models/imported_item.rb
```diff
@@ -18,7 +18,7 @@ def import(params)
       contacts << Contact.new(
         test_and_trace_account_id: row[0],
         nhs_number: row[1],
-        is_vulnerable: (row[2] == 1 || row[2]&.downcase == 'true'),
+        is_vulnerable: (row[2] == 1 || row[2].to_s&.downcase == 'true'),
```
go8soft, line 21: good catch, thanks. We can remove `&` in this case.  It should be:
`is_vulnerable: (row[2] == 1 || row[2].to_s.downcase == 'true'),`

userman123: makes sense 👍

PR #360, opened 2020-12-04
Add nhs number unique validation

go8soft:
Please note we have unique migration and if we have not unique values on production it will fail.

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -14,6 +14,7 @@ class Contact < ApplicationRecord
   has_paper_trail
 
   validates :first_name, presence: true
+  validates :nhs_number, uniqueness: true
   validates_date :date_of_birth, allow_nil: true, allow_blank: true
 
   scope :search, lambda { |text|
--- a/db/migrate/20201204164913_add_contacts_nhs_number_unique.rb
+++ b/db/migrate/20201204164913_add_contacts_nhs_number_unique.rb
@@ -1,0 +1,6 @@
+class AddContactsNhsNumberUnique < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+  def change
+    add_index :contacts, :nhs_number, unique: true, algorithm: :concurrently
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_11_19_125747) do
+ActiveRecord::Schema.define(version: 2020_12_04_164913) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -48,6 +48,7 @@ ActiveRecord::Schema.define(version: 2020_11_19_125747) do
     t.string "lead_service_note"
     t.bigint "imported_item_id"
     t.string "test_and_trace_account_id"
+    t.index ["nhs_number"], name: "index_contacts_on_nhs_number", unique: true
   end
 
   create_table "imported_items", force: :cascade do |t|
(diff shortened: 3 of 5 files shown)


db/migrate/20201204164913_add_contacts_nhs_number_unique.rb
```diff
@@ -0,0 +1,11 @@
+class AddContactsNhsNumberUnique < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def up
+    safety_assured { execute 'create unique index index_contacts_on_lower_nhs_number on contacts using btree (lower(nhs_number))' }
+  end
+
+  def dow
```
userman123, line 8: is this a typo? should be 'down'?

app/models/contact.rb
```diff
@@ -14,6 +14,7 @@ class Contact < ApplicationRecord
   has_paper_trail
 
   validates :first_name, presence: true
+  validates :nhs_number, uniqueness: { case_sensitive: false }
```
userman123, line 17: think this needs `, allow_blank: true` added otherwise it will think its a duplicate when blank

features/step_definitions/assessments/fail_assessment_steps.rb
```diff
@@ -1,5 +1,5 @@
 Given('an assessment is assigned to me') do
-  @contact = Contact.create!(first_name: 'Test')
+  @contact = Contact.create!(first_name: 'Test', nhs_number: Time.now.to_f)
```
userman123, line 2: I'm guessing this was added because of the allow blank issue mentioned above?

features/step_definitions/permissions/permissions_search_steps.rb
```diff
@@ -1,5 +1,5 @@
 And('another contact {string} is not visible to me') do |name|
-  @another_contact = Contact.create!(first_name: name + rand(10**10).to_s(36))
+  @another_contact = Contact.create!(first_name: name + rand(10**10).to_s(36), nhs_number: Time.now.to_f)
```
userman123, line 2: this also - nhs_number should not be required

db/migrate/20201204164913_add_contacts_nhs_number_unique.rb
```diff
@@ -0,0 +1,11 @@
+class AddContactsNhsNumberUnique < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def up
+    safety_assured { execute 'create unique index index_contacts_on_lower_nhs_number on contacts using btree (lower(nhs_number))' }
```
userman123, line 5: I don't think this will work as there are already duplicates in the database. Not sure what best way forward is - maybe we just add validations at code level rather than database?

go8soft: Yes, we can add only model validations. But we will not be able to edit contacts with duplicate `nhs_number`. Can we:
1 - validate we have there are already duplicates in the database on production
2 - if yes, to ask what to do with them
?

db/migrate/20201204164913_add_contacts_nhs_number_unique.rb
go8soft: Yes, thanks, my bad. Sorry about that. Will fix it.

db/migrate/20201204164913_add_contacts_nhs_number_unique.rb
dmgarland: If there are duplicates for the same contact could you get away with adding a scope? https://guides.rubyonrails.org/active_record_validations.html#uniqueness

go8soft: Thanks @dmgarland we can use it for DOB

userman123:
This is still waiting on duplicates being resolved on prod

PR #363, opened 2020-12-10
Strip whitespace from all AR attributes that respond_to strip - Contacts model

WorkSutton:
This is to prevent future duplicates entering the system via user input on the `Add person form`

`Mark     ` will become `Mark`

A separate PR will be rasied to clean existing data

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -5,7 +5,7 @@ class Contact < ApplicationRecord
   require 'activerecord-import/active_record/adapters/postgresql_adapter'
   include CleanData
 
-  before_save :strip_whitespace
+  before_validation :strip_whitespace_from_all_text_and_strings
 
   has_many :needs, dependent: :destroy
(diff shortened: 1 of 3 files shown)


go8soft (approved):

app/models/contact.rb
```diff
@@ -3,6 +3,9 @@
 class Contact < ApplicationRecord
   require 'activerecord-import/base'
   require 'activerecord-import/active_record/adapters/postgresql_adapter'
+  include CleanData
+
+  before_save :strip_whitespace
```
dmgarland, line 8: Maybe `before_validate` ? in case that affects some validation rule

WorkSutton: @dmgarland Thats a fair point, I had not thought of that. I will update and put back for review.

Additionally do we want to specify which fields we want to target?  This could be then called with `strip_whitespace <field1> <field2> etc

PR #364, opened 2020-12-10
Strip off leading/trailing whitespace from imports

WorkSutton:
Clean up imports from spreadsheets

### DIFF
--- a/app/models/imported_item.rb
+++ b/app/models/imported_item.rb
@@ -22,17 +22,17 @@ class ImportedItem < ApplicationRecord
     contacts = []
     unique_contacts.each do |row|
       contacts << Contact.new(
-        test_and_trace_account_id: row[0],
-        nhs_number: row[1],
+        test_and_trace_account_id: row[0].strip,
+        nhs_number: row[1].strip,
         is_vulnerable: (row[2] == 1 || row[2].to_s.downcase == 'true'),
-        first_name: row[3],
-        surname: row[4],
+        first_name: row[3].strip,
+        surname: row[4].strip,
         date_of_birth: row[5],
-        email: row[6],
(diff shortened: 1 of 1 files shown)


go8soft:
@WorkSutton we dont need this PR, because of https://github.com/wearefuturegov/beacon/pull/36 and https://apidock.com/rails/ActiveRecord/Callbacks/before_save. Because we are using Contact object it will take care for the strip.

WorkSutton:
Ah good catch @go8soft. I had created this first but didn't push it up. I wanted this as a belt and braces incase we did not go ahead with the other PR.

app/models/imported_item.rb
```diff
@@ -22,17 +22,17 @@ def import(params)
...
+        nhs_number: row[1].strip,
         is_vulnerable: (row[2] == 1 || row[2].to_s.downcase == 'true'),
-        first_name: row[3],
-        surname: row[4],
+        first_name: row[3].strip,
+        surname: row[4].strip,
         date_of_birth: row[5],
-        email: row[6],
+        email: row[6].strip,
         mobile: row[7],
```
dmgarland, line 32: Why are some fields stripped and not others?

WorkSutton: @dmgarland I targeted text/string fields. Basically ones that will respond to strip

WorkSutton:
Given @go8soft has correctly identified that if my other [PR 363](https://github.com/wearefuturegov/beacon/pull/363) is approved by all, and it will mean that imported data is saved correctly too, should we leave this as a belt and braces, or just drop it?

WorkSutton:
Closed as functionality will be delivered in PR363

PR #366, opened 2020-12-14
Run rake task to strip whitespace from existing contacts

WorkSutton:
The PR can be used to strip whitespace from any table/model within the Beacon app.

### How to strip whitespace from a model

The core functionality was extracted from the original rake task (not committed) written to solve the Contacts table whitespace search issue. This is now located in the `HistoricDataCleaner` class. 
The object of interest calls `save` which will update all records that are returned in the `#process_in_batches`.

The save method is dependant on the model concern `stripable`.

The class can be either called directly, or as is proposed from a Rake task.

### Rake Task

`rake -T | grep strip`

`rake strip_whitespace:data_cleanup[object_name,first_attribute]  # Strip whitespace from specified model/table`

The task is namespaced (strip_whitespace) and is called `data_cleanup`. It takes two arguments:
- table/model name (singular)
- table fields/model.attribute_names of the string/text fields to be updated

Example rake command:

`RAILS_ENV=<target_environment> bundle exec rake strip_whitespace:data_cleanup[<model/object>,<field/attribute_1>]` at a minimum.

```
RAILS_ENV=development bundle exec rake strip_whitespace:data_cleanup[contacts,first_name,middle_names,surname]
```

### Output (STDOUT/Log file)

Two output streams are broadcast to as a result of the call to `#create_logger`

A log file is generated named as follows:

`log/<caller>_<object_name>_<Rails.env>.log`

- `caller` is the name of the calling program - for the rake task it is `namespace_taskname`
- `object_name` is the name of the model/table (singular) that is the target of the operation
- `Rails.env` - hopefully self explanatory

In addition to a log file being generated, logging also is sent to STDOUT (terminal/console)

### Successful Run
A successful run will generate output as per the screenshot below.

STDOUT-successful-run-with-two-errors

In this run there are two records which have failed, marked with a red `X` during the run and in the generated report.
Both the records have the ID and reason displayed in teh `Error Records ID & reason` section.

To test, I had an additional rake task to update some data. See section `Non commited rake task (data setup)`

### Possible Errors (from rake task)
There are several scenarios whereby errors may be encountered. Each of these are documented with screenshots showing what caused them and the output (terminal screenshots)

#### Passing an incorrect object/table name

STDOUT-invalid-table-object-name-passed

#### Passing incorrect field/attribute names

STDOUT-invalid-field-or-attribute-names-passed

### Alternate environments

At the top of the output clearly shows the environment targeted.

STDOUT-test-environment

STDOUT-production-environment (mocked with development)

#### Example log file output

strip_whitespace_data_cleanup_contact_development log

### Non-commited Rake task

The following rake task was used to test/develop this feature

```ruby
    task backup_setup_data: :environment do
      ActiveRecord::Base.connection.execute(
      "
      UPDATE contacts SET first_name = '',     middle_names = 'Lana',                 surname = '  Bergstrom   '                WHERE id = 98;
      UPDATE contacts SET first_name = '',     middle_names = 'Gerlach',              surname = ' Nikolaus '                    WHERE id = 99;

      UPDATE contacts SET first_name = 'Laurel',         middle_names = 'Windler',              surname = 'Brekke'              WHERE id = 100;
      UPDATE contacts SET first_name = 'Noah',           middle_names = 'Kris',                 surname = 'Kuvalis    '         WHERE id = 101;
      UPDATE contacts SET first_name = '    Demetrius ', middle_names = 'Ryan',                 surname = '          Barrows  ' WHERE id = 102;
      UPDATE contacts SET first_name = '    Shirley ',   middle_names = 'Cassin',               surname = '     Ward    '       WHERE id = 103;
      UPDATE contacts SET first_name = '    Rickie ',    middle_names = 'Vandervort',           surname = 'Lind    '            WHERE id = 104;
      UPDATE contacts SET first_name = ' Mckinley   ',   middle_names = 'Pfannerstill',         surname = ' Pagac'              WHERE id = 105;
      UPDATE contacts SET first_name = ' Kathlene   ',   middle_names = 'Zieme',                surname = '     Johnston'       WHERE id = 106;
      UPDATE contacts SET first_name = ' Refugio   ',    middle_names = 'Bergstrom',            surname = '  Casper           ' WHERE id = 107;
      UPDATE contacts SET first_name = '  Keren    ',    middle_names = 'Heaney',               surname = '   Greenholt '       WHERE id = 108;
      UPDATE contacts SET first_name = '   Carley    ',  middle_names = 'Muller',               surname = '   Effertz  '        WHERE id = 109;
      UPDATE contacts SET first_name = '  Arlyne',       middle_names = 'Swaniawski',           surname = '  Wolf  '            WHERE id = 110;
      UPDATE contacts SET first_name = '   Elna ',       middle_names = 'Bernier',              surname = 'Cassin '             WHERE id = 111;
      UPDATE contacts SET first_name = '   Barney  ',    middle_names = 'Berge',                surname = '  Shanahan'          WHERE id = 112;
      UPDATE contacts SET first_name = '   Stan   ',     middle_names = 'Baumbach',             surname = '  Terry   '          WHERE id = 113;
      UPDATE contacts SET first_name = '  Missy    ',    middle_names = 'Cronin',               surname = ' Olson '             WHERE id = 114;
      UPDATE contacts SET first_name = ' Timmy',         middle_names = 'Becker',               surname = '   Hackett'          WHERE id = 115;
      UPDATE contacts SET first_name = ' Jeneva',        middle_names = 'Wuckert',              surname = '  OConnell'          WHERE id = 116;
      UPDATE contacts SET first_name = '   Johnnie',     middle_names = 'Ankunding',            surname = ' Macejkovic   '      WHERE id = 117;
      UPDATE contacts SET first_name = '    Julius',     middle_names = 'Bauch',                surname = 'Jakubowski  '        WHERE id = 118;
      UPDATE contacts SET first_name = 'Marketta  ',     middle_names = 'Harris',               surname = 'Dach        '        WHERE id = 119;
      UPDATE contacts SET first_name = 'Arnulfo  ',      middle_names = 'Breitenberg',          surname = 'Littel '             WHERE id = 120;
      UPDATE contacts SET first_name = 'Leigha ',        middle_names = 'Legros',               surname = '    Rohan'           WHERE id = 121;
      UPDATE contacts SET first_name = ' Jarred ',       middle_names = 'Cormier',              surname = '    Mraz   '         WHERE id = 122;
      UPDATE contacts SET first_name = 'Laurence  ',     middle_names = 'Schmeler',             surname = 'Herman '             WHERE id = 123;
      UPDATE contacts SET first_name = 'Katy   ',        middle_names = 'Turcotte',             surname = ' Buckridge'          WHERE id = 124;
      UPDATE contacts SET first_name = 'Donovan   ',     middle_names = 'Metz',                 surname = '  Schiller'          WHERE id = 125;
      UPDATE contacts SET first_name = 'Trinity    ',    middle_names = 'Schmeler',             surname = '  Abbott   '         WHERE id = 126;
      UPDATE contacts SET first_name = '   Kelli',       middle_names = 'Block',                surname = 'Ondricka   '         WHERE id = 127;
      UPDATE contacts SET first_name = '  Olivia',       middle_names = 'Wilkinson',            surname = 'Schaden  '           WHERE id = 128;
      UPDATE contacts SET first_name = ' Claudia ',      middle_names = 'Schuster',             surname = 'Crist   '            WHERE id = 129;
      UPDATE contacts SET first_name = 'Tiffanie  ',     middle_names = 'Metz',                 surname = 'Macejkovic'          WHERE id = 130;
      UPDATE contacts SET first_name = '  Guy ',         middle_names = 'Jaskolski',            surname = '    Leannon'         WHERE id = 131;
      UPDATE contacts SET first_name = '    Irving    ', middle_names = 'Klein',                surname = '  Brown     '        WHERE id = 132;
      UPDATE contacts SET first_name = 'Seth',           middle_names = 'Walter',               surname = 'Daugherty   '        WHERE id = 133;
      UPDATE contacts SET first_name = '    Danilo',     middle_names = 'Gerlach',              surname = ' Nikolaus '          WHERE id = 134;
      "
      ) # 35 records
    end
```

Embeded as an additonal task in the existing namespace.

### DIFF
--- a/lib/historic_data_cleaner.rb
+++ b/lib/historic_data_cleaner.rb
@@ -1,0 +1,165 @@
+# frozen_string_literal: true
+
+class HistoricDataCleaner
+  BATCH_SIZE = 5
+  DOUBLE_LINE = '================================================='
+  FAILURE = '❌'
+  LINE = '------------------------------------'
+  PASS = '✅'
+
+  def self.call(object, caller, fields_to_update)
+    new(object, caller, fields_to_update)
+  end
+
+  attr_reader :multi_logger
+
+  def initialize(object, caller, fields_to_update)
+    @caller = caller
+    @environment_log_message = environment_log_message
+    @fields_to_update = fields_to_update
+    @object_name = object
+    @multi_logger = create_logger
+    @object_statement = []
+    @updated_records = []
+    @invalid_records = []
+    @processed_count = 0
+    @updated_count = 0
+    @not_updated_count = 0
+  end
+
+  def cleanse
+    log_start
+    validate_object
+    build_object_statement
+    validate_object_statement
+    process_in_batches
+    report_success_and_errors
+  end
+
+  private
+
+  attr_accessor :object_statement, :updated_records, :invalid_records, :processed_count, :updated_count, :not_updated_count
+  attr_reader :fields_to_update, :object
+
+  def caller=(caller)
+    @caller = caller.include?(':') ? caller.split(':').join('-') : caller
+  end
+
+  def environment_log_message
+    case Rails.env
+    when 'development'
+      "\e[1m\e[32m#{Rails.env.upcase}\e[0m"
+    when 'test'
+      "\e[36m#{Rails.env.upcase}\e[0m"
+    when 'production'
+      "\e[31mCAUTION: \u001b[4m\e[0;101m\e[1;97mPRODUCTION\e[0m"
+    else
+      "\e[1;33mWARN: \e[1m\e[1;33m#{Rails.env.upcase}\e[0m"
+    end
+  end
+
+  def create_logger
+    stdout_logger = ActiveSupport::Logger.new(STDOUT)
+    file_logger = ActiveSupport::Logger.new("log/#{@caller}_#{@object_name}_#{Rails.env}.log")
+    multi_logger = stdout_logger.extend(ActiveSupport::Logger.broadcast(file_logger))
+    multi_logger.level = Logger::INFO
+
+    original_formatter = Logger::Formatter.new
+    file_logger.formatter = proc { |severity, datetime, progname, msg|
+      original_formatter.call(severity, datetime, progname, msg)
+    }
+    multi_logger
+  end
+
+  def log_start
+    log "\u001b[0m\u001b[7m ENVIRONMENT ** \e[0m #{environment_log_message} \u001b[0m\u001b[7m ** ENVIRONMENT \e[0m\n"
+    log "#{Time.now}  \e[1m\e[33m__START__\e[0m"
+    msg = "Processing Data Cleanup Request for #{@object_name.capitalize}s"
+    log msg
+    puts DOUBLE_LINE
+  end
+
+  def log(msg, status = :info)
+    severity = Object.const_get("#{Logger}::#{status.to_s.upcase}")
+    multi_logger.add(severity, msg, self.class.name)
+  end
+
+  def validate_object
+    @object = @object_name.to_s.classify.constantize
+  rescue NameError => e
+    log("#{e.class}: #{e.message}", :fatal)
+    log("\e[31mPlease check the name of the model/table is correct - Exiting\e[0m")
+    exit 90
+  end
+
+  def build_object_statement
+    log "Fields/columns/attributes - targeted\n#{LINE}"
+    fields_to_update.each do |field|
+      log field
+      object_statement << "LENGTH(RTRIM(LTRIM(#{field}))) <> LENGTH(#{field})"
+    end
+    puts LINE + "\n"
+  end
+
+  def validate_object_statement
+    valid_fields = object.attribute_names.select { |name| [:string, :text].include?(object.type_for_attribute(name).type) }
+    invalid_fields = fields_to_update - valid_fields
+    return unless invalid_fields.any?
+
+    log('ActiveRecord::InvalidStatement would be raised', :fatal)
+    invalid_fields.each do |field|
+      log("column \e[31m#{@object_name}.#{field}\e[0m does not exist")
+    end
+    log("\e[31mPlease check the name of the fields are correct - Exiting\e[0m")
+    exit 95
+  end
+
+  def process_in_batches
+    temp_updated_records = []
+    log "BATCH_SIZE: #{BATCH_SIZE}\n#{LINE}"
+
+    # Process in batches set by the batch size 1000 is default.
+    object
+      .where(object_statement.join(' OR '))
+      .find_in_batches(batch_size: BATCH_SIZE)
+      .with_index do |group, batch|
+        temp_updated_records = []
+        msg = "\nProcessing batch #{batch} (#{group.size} records)"
+        print msg
+        msg = '.'
+
+        group.each do |record|
+          if record.save(touch: false)
+            @updated_count += 1
+            temp_updated_records << record.id
+            print "\e[32m#{msg}\e[0m"
+          else
+            @not_updated_count += 1
+            @invalid_records << "#{record.id} - #{record.errors.full_messages.to_sentence}"
+            print "\e[31mx\e[0m"
+          end
+          @processed_count += 1
+        end
+        @updated_records << temp_updated_records
+      end
+  end
+
+  def report_success_and_errors
+    if updated_records.empty?
+      log "\nZERO records!\n"
+    else
+      log "\n#{LINE}\nTotal Records Processed: #{processed_count}"
+      log "Updated #{PASS} #{updated_count}"
+
+      if not_updated_count.positive?
+        log "Errors #{FAILURE} #{not_updated_count}"
+        log "#{LINE}\nError Records ID & reason\n#{invalid_records.split.join("\n")}"
+      end
+      unless updated_records.flatten.empty?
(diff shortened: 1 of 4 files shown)


dmgarland:
Amazing write-up Mark, really clear explanation.

I wasn't on the initial conversation so I've got some possibly moot questions:

1) How does this problem occur in the first place? Has that been addressed?
2) If so, is this rake task a one-off fix?
3) ... because if it is it's quite a bit of code to add to solve this problem

Have we considered the approach of just doing this in SQL as a one-off dev task and fixing whatever validation / data scrubbing is needed. I like the implementation but I thought I'd just throw out there the flip side is extra code = extra maintenance etc 

Since it's finished this point is badly timed so sorry in advance, perhaps it makes sense to merge it to solve the problem in the short term?

lib/historic_data_cleaner.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class HistoricDataCleaner
-  BATCH_SIZE = 1000
+  BATCH_SIZE = Rails.env.development? || Rails.env.test? ? 5 : 1000
```
dmgarland, line 4: Personally am not a fan of environment conditional code, I'd sooner use an environment variable here rather than having logic if you need to override it; but in practice I think you can keep the constant of 1000 and then stub the value in your test if you want to see 5 at a time.

WorkSutton: @dmgarland Thats a great point about stubbing it in the test. I had not thought about that as I didn't write any tests for this.

Thanks, that's a good point about using environment variable vs conditional

WorkSutton:
@dmgarland Thanks for the review and comments, plus positive feedback on the writeup.

I attach my responses to your questions.

> 1. How does this problem occur in the first place? Has that been addressed?
The problem occurred because whitespace was not being stripped from data previously entered in the system - either manually or via imports. 
A fixed in [PR-363](https://github.com/wearefuturegov/beacon/pull/363) added a concern containing a module which is called in a before_validation callback. That then prevents any future whitespace issues for string/text fields when the concern is included in a model.

> 2. If so, is this rake task a one-off fix?
Technically the ticket is for a one off task, but I see no reason why the issue does not exist elsewhere and it was raised by Enna at Camden as an issue in search results.

So I reasoned - possibly incorrectly - that we have other models to clean, and built it so it was extensible. 
In retrospect it was possibly overkill.

> 3. ... because if it is it's quite a bit of code to add to solve this problem
Acknowledged!
> 
> Have we considered the approach of just doing this in SQL as a one-off dev task and fixing whatever validation / data scrubbing is needed. I like the implementation but I thought I'd just throw out there the flip side is extra code = extra maintenance etc

Fair comment on maintenance.

I'm not a huge fan of rake tasks for data cleanup, I would rather write a SQL query but too many places I have worked have drummed it into me to simply use the Ruby tools for the task, Or they won't even consider a SQL query. So I have kind of gotten to a point where I just default to writing a rake task.

Other option was a DATA migration - either as a standalone migration using one of the specific gems for this ( but we are not using) or within a regular rails migration. But I am not a fan of writing any DML (Data Manipulation Language) / TCL (Transaction Control Language) queries in default Rails migrations. I want them to be pure DDL (Data Definition Language) or DCL (Data Control Language).

dmgarland:
So in response to your points:

1. In the case of manual inputs, whatever happens with this PR we ought to address the problem at source too - validation client/server side. For imports I'd say it's the responsibility of the importer to strip any space

2. If it's solving a generic problem that's fine but it lives in a gem rather than in this codebase

3. IMO making one-off changes sounds like a SQL script that an admin can run and is separate to the development process with all the agile, testing what not that goes with it. YMMV. I reckon the downside is lots of verbose rake tasks that no-one really knows what they do and may or may not be generic enough to solve all future problems anyway.

WorkSutton:
@dmgarland 

> 1. In the case of manual inputs, whatever happens with this PR we ought to address the problem at source too - validation client/server side. For imports I'd say it's the responsibility of the importer to strip any space

Yes client side - I think you mean in JS? is a good suggestion too

Agree on the importer side. In this instance are we saying the `CleanData` concern is the wrong place? As that handles any data thrown at the model that implements its methods. Or is that overkill in your opinion too?

> 2. If it's solving a generic problem that's fine but it lives in a gem rather than in this codebase

Excellent point. I didn't do that due to time issues, if I had more time I would have asked if we wanted just for project or in  gem for other projects to reuse.

> 3. IMO making one-off changes sounds like a SQL script that an admin can run and is separate to the development process with all the agile, testing what not that goes with it. YMMV. I reckon the downside is lots of verbose rake tasks that no-one really knows what they do and may or may not be generic enough to solve all future problems anyway.

I concur on your point on rake tasks. I have seen that on many projects over the years. They get added to do these tasks and usually never get removed. I make a point of removing them if the task is true data cleanup and one/few times only.

Yes the overhead of doing a rake task to me often seems overkill when a simple SQL script would suffice.

userman123:
So for reference, @go8soft , @WorkSutton  and I discussed this last night and it seems like really great work but unfortunately is probably too complex for our requirements - don't have time to fully review and be confident in what it does. Appreciate the reusability factor but for now its unlikely to be needed anywhere else. @WorkSutton is going to make a simplified version

WorkSutton:
Closing - simplified version supplied [PR-372](https://github.com/wearefuturegov/beacon/pull/372)

PR #367, opened 2020-12-14
Add assign_role_if_empty for imported item policy

go8soft:


WorkSutton (approved):
LGTM @go8soft

app/policies/imported_item_policy.rb
```diff
@@ -1,6 +1,7 @@
 class ImportedItemPolicy < ApplicationPolicy
   class Scope < Scope
     def resolve
+      @user.assign_role_if_empty
```
userman123, line 4: Not sure I understand why we need to do this? But it does fix the bug 👍

PR #369, opened 2020-12-15
Feature/style guide

dmgarland:
Enough plumbing to bring in the style guide


userman123:
Can run locally, new styles look good. However some things not quite right, like when I hover over an imported item there is no highlight or cursor change - not clear you can click on it. Also the file uploader seems to be broken. Please see screenshots. Can't see rejected contacts either, guessing that's what you are doing next? Could you rebase onto master please also as I think there are gonna be a lot of conflicts with Mark's work that's already merged in.

Before:
Screenshot 2020-12-17 at 13 28 14

Now:
Screenshot 2020-12-17 at 13 28 41
Screenshot 2020-12-17 at 13 28 45

File does not appear to attach after selecting:
Screenshot 2020-12-17 at 13 30 21
Screenshot 2020-12-17 at 13 30 31

dmgarland:
Thanks for the catch @userman123. I was using `DOMContentLoaded` instead of `turbolinks:load` - keep forgetting this is a turbolinks app!

dmgarland:
Regarding hover states, they aren't in the new design so I'll check with Nick to see whether we're keeping them

dmgarland:
And finally regarding the conflicts, I think I'll just drop Marks' last commit because I've implemented the new designs with those fields anyway

WorkSutton:
> And finally regarding the conflicts, I think I'll just drop Marks' last commit because I've implemented the new designs with those fields anyway

Yeap @dmgarland thats the best way forward as you have covered this off in your work here. Sorry for causing the overlap

WorkSutton (approved):
LGTM @dmgarland

PR #370, opened 2020-12-15
Add 2 additional fields to data import

WorkSutton:
Add additional fields from import spreadsheet

- `test_trace_creation_date`
- `isolation_start_date`

Some minor refactoring to names in import method as `row` did not feel right in all places referenced

Move Rails.logger.unknown message about `User imported new contacts` to after Rollback, as if Rollback occurs then technically zero new records are imported

```ruby
[1] pry(main)> Contact.last
  Contact Load (0.4ms)  SELECT "contacts".* FROM "contacts" ORDER BY "contacts"."id" DESC LIMIT $1  [["LIMIT", 1]]
=> #<Contact:0x00007f7878a037d0
 id: 639,
 first_name: "Casey",
 middle_names: nil,
 surname: "Addams",
 address: "201 White Shires Road",
 postcode: "N1 0AB",
 telephone: nil,
 mobile: "+44762",
 gds_import_data: nil,
 is_vulnerable: false,
 created_at: Tue, 15 Dec 2020 15:45:45 UTC +00:00,
 updated_at: Tue, 15 Dec 2020 15:45:45 UTC +00:00,
 email: nil,
 additional_info: nil,
 count_people_in_house: nil,
 any_children_under_age: nil,
 delivery_details: nil,
 any_dietary_requirements: nil,
 dietary_details: nil,
 cooking_facilities: nil,
 eligible_for_free_prescriptions: nil,
 nhs_number: "100000000",
 date_of_birth: Mon, 15 Feb 1971,
 has_covid_symptoms: nil,
 lock_version: 0,
 channel: nil,
 no_calls_flag: false,
 deceased_flag: false,
 share_data_flag: nil,
 lead_service_id: nil,
 lead_service_note: nil,
 imported_item_id: 44,
 test_and_trace_account_id: "yyddmm20201215",
 test_trace_creation_date: Thu, 03 Dec 2020,
 isolation_start_date: nil>
[2] pry(main)> 

```

mocked-up-data

The last row in `png` is the record above.

### DIFF
--- a/app/models/imported_item.rb
+++ b/app/models/imported_item.rb
@@ -17,25 +17,25 @@ class ImportedItem < ApplicationRecord
       contacts_rows << row
     end
 
-    unique_contacts, not_unique_contacts = unique_and_non_unique_records contacts_rows
+    unique_contacts, not_unique_contacts = unique_records contacts_rows
 
     contacts = []
-    unique_contacts.each do |contact_record|
+    unique_contacts.each do |row|
       contacts << Contact.new(
-        test_and_trace_account_id: contact_record[0],
-        nhs_number: contact_record[1],
-        is_vulnerable: (contact_record[2] == 1 || contact_record[2].to_s.downcase == 'true'),
-        first_name: contact_record[3],
-        surname: contact_record[4],
-        date_of_birth: contact_record[5],
-        email: contact_record[6],
-        mobile: contact_record[7],
-        telephone: contact_record[8],
-        address: contact_record[9],
-        postcode: contact_record[10],
-        needs: [Need.new(category: 'Check in', start_on: Date.today + 2.days, name: contact_record[11])],
-        test_trace_creation_date: contact_record[12],
-        isolation_start_date: contact_record[13],
+        test_and_trace_account_id: row[0],
+        nhs_number: row[1],
+        is_vulnerable: (row[2] == 1 || row[2].to_s.downcase == 'true'),
+        first_name: row[3],
+        surname: row[4],
+        date_of_birth: row[5],
+        email: row[6],
+        mobile: row[7],
+        telephone: row[8],
+        address: row[9],
+        postcode: row[10],
+        needs: [Need.new(category: 'Check in', start_on: Date.today + 2.days, name: row[11])],
+        test_trace_creation_date: row[12],
+        isolation_start_date: row[13],
         imported_item: self
       )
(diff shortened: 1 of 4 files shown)


WorkSutton:
Have not added the two fields to the permitted params for Contacts. As not yet added to views

app/models/imported_item.rb
```diff
@@ -17,25 +17,25 @@ def import(params)
       contacts_rows << row
     end
 
-    unique_contacts, not_unique_contacts = unique_records contacts_rows
+    unique_contacts, not_unique_contacts = unique_and_non_unique_records contacts_rows
```
userman123, line 20: Why is this now unique_and_non_unique? Was it wrong before?

Not sure I ever understood the syntax on this line actually, what does it mean to have `x, y = z`

app/models/imported_item.rb
```diff
@@ -17,25 +17,25 @@ def import(params)
...
-        date_of_birth: row[5],
-        email: row[6],
-        mobile: row[7],
-        telephone: row[8],
-        address: row[9],
-        postcode: row[10],
-        needs: [Need.new(category: 'Check in', start_on: Date.today + 2.days, name: row[11])],
-        test_trace_creation_date: row[12],
-        isolation_start_date: row[13],
+        test_and_trace_account_id: contact_record[0],
```
userman123, line 25: not 100% sure contact_record is clearer than row - row makes it clear it is coming from spreadsheet - but not a big deal I don't think

app/models/imported_item.rb
userman123: is it assigning both unique_contacts and not_unique_contacts the same thing?

userman123: ah just looked up think I understand now. So `unique_and_non_unique_records contacts_rows` returns `[unique_contacts, not_unique_contacts]` and then each get assigned accordingly. Makes sense I think

userman123: renaming of method makes sense in that case :)

userman123:
Happy with everything apart from the refactoring - I think calling everything a record is confusing. But appreciate it's not crystal clear as is.  Can you move that commit ( #44d0a9cfb10b15106bf3c6070cdb6e1884f80ab2) a separate PR please so we can review later with @go8soft

userman123 (approved):

PR #371, opened 2020-12-16
Record duplicates that failed to import

WorkSutton:
Record all records which fail validation from importing contacts as RejectedContact

Add some information to ImportedItem for:
- number of records successfully imported
-  number of records rejected
- who requested the import


db/migrate/20201215191958_create_rejected_contacts.rb
```diff
@@ -0,0 +1,22 @@
...
+      t.string :first_name
+      t.string :middle_names
+      t.string :surname
+      t.date :date_of_birth
+      t.text :email
+      t.string :mobile
+      t.string :telephone
+      t.string :address
+      t.string :postcode
+      t.string :needs
```
userman123, line 17: what's `needs` column for?

db/migrate/20201216163610_add_import_count_and_failed_count_and_userid_to_imported_items.rb
```diff
@@ -0,0 +1,7 @@
+class AddImportCountAndFailedCountAndUseridToImportedItems < ActiveRecord::Migration[6.0]
+  def change
+    add_column :imported_items, :imported, :integer
+    add_column :imported_items, :failed, :integer
+    add_column :imported_items, :uploaded_by_user_id, :bigint
```
go8soft, line 5: Please create fk here and use `user_id` instead of `uploaded_by_user_id`.

go8soft, line 5: Please create fk here and use user_id instead of uploaded_by_user_id.

db/migrate/20201215191958_create_rejected_contacts.rb
```diff
@@ -0,0 +1,22 @@
+class CreateRejectedContacts < ActiveRecord::Migration[6.0]
+  def change
```
go8soft, line 2: We need only fields from the import document + id + timestamps + imported item id + reason

db/migrate/20201215191958_create_rejected_contacts.rb
userman123: ah I see, so for an imported contact it creates an actual need, but for a reject, we just record it here? Makes sense I think. @go8soft thoughts?

WorkSutton: @userman123 
Because its in the import so it just captures the data that was asked to be loaded. As it is rejected we cannot create a need at this point in time

app/models/imported_item.rb
```diff
@@ -60,10 +93,13 @@ def unique_records(rows)
 
   def check_row(row)
     if not_empty(row[0]) && not_zero(Contact.where('test_and_trace_account_id = ?', row[0].to_s))
+      row << "Test and Trace Account ID - Duplicate [#{row[0]}]"
       row
```
go8soft, line 97: You can remove `row`, because `row << ...` will return the array

go8soft (changes requested):
Please adapt also https://github.com/wearefuturegov/beacon/blob/6907d9dfc426fa2934bacca34df664c140c15f50/app/controllers/imported_items_controller.rb#L42 message

db/migrate/20201215191958_create_rejected_contacts.rb
go8soft: @userman123 yes I wrote them in the review

db/migrate/20201216163610_add_import_count_and_failed_count_and_userid_to_imported_items.rb
WorkSutton: In a chat elsewhere it was suggested to name the record this way. Have actioned your request to change.

db/migrate/20201215191958_create_rejected_contacts.rb
WorkSutton: @go8soft As discussed on our call there were additional fields. Not sure about one name so will check and alter if required

db/migrate/20201216184415_create_rejected_contacts.rb
```diff
@@ -2,11 +2,11 @@ class CreateRejectedContacts < ActiveRecord::Migration[6.0]
   def change
     create_table :rejected_contacts do |t|
       t.references :imported_item, null: false, foreign_key: true
+      t.string :reason
       t.string :test_and_trace_account_id
       t.string :nhs_number
       t.boolean :is_vulnerable
       t.string :first_name
-      t.string :middle_names
```
userman123: Shouldn't really be changing migrations after pushing as it breaks database for anyone who's already run - should be a new migration ideally. But given it's not been deployed anywhere, not the end of the world - I can just reset my local database.

db/migrate/20201216184644_add_import_count_and_failed_count_and_userid_to_imported_items.rb
```diff
@@ -1,7 +1,8 @@
 class AddImportCountAndFailedCountAndUseridToImportedItems < ActiveRecord::Migration[6.0]
   def change
     add_column :imported_items, :imported, :integer
-    add_column :imported_items, :failed, :integer
-    add_column :imported_items, :uploaded_by_user_id, :bigint
+    add_column :imported_items, :rejected, :integer
   end
+
+  add_reference :imported_items, :user, foreign_key: true, index: { algorithm: :concurrently }
```
userman123, line 7: this probably just my ignorance, buy why does the `add_reference` go outside of the `def change`?

userman123 (approved):

db/migrate/20201216184415_create_rejected_contacts.rb
WorkSutton: Agreed @userman123, I just went for fastest solution. Would totally not have done this if it had been deployed anywhere

db/migrate/20201216184644_add_import_count_and_failed_count_and_userid_to_imported_items.rb
WorkSutton: Its to do with strong migrations, well at least I ran into issues with it telling me what to do.

PR #372, opened 2020-12-16
Remove historic whitespace from strings/text in Contacts data

WorkSutton:
[Trello 86 - whitespace removal - CONTACTS](https://trello.com/c/88Ps7pt9/86-data-cleanup-strip-whitespace-from-file-provided-for-import)

Simplified version.

Will close [PR 366](https://github.com/wearefuturegov/beacon/pull/366)


lib/tasks/historic_data_contacts_whitespace_strip.rake
```diff
@@ -0,0 +1,13 @@
+namespace :strip_whitespace do
+  desc 'Strip whitespace from specified historic Contact entrys (one off task)'
+  task data_cleanup: :environment do
+    BATCH_SIZE = 1000
+    # Process in batches set by the batch size 1000 is default.
+    Contact.find_in_batches(batch_size: BATCH_SIZE) do |contact|
+      puts contact.size
```
go8soft, line 7: If we have 5 100 records we will have:
1000
1000
1000
1000
100

We can just have some ids output like:
12,13,14,15....
to show the progress

go8soft (approved):
LGTM :+1: minor output log comment, but not mandatory
@userman123 please double check we have backup before exec on prod

lib/tasks/historic_data_contacts_whitespace_strip.rake
WorkSutton: There was a progress meter in the previous one. I just made this super simple. Maybe too simple. @go8soft what about failures? There should be none, but how do you propose to display those?

This was why in the closed PR I wrote a sql statement to execute so we knew that only records that needed updating are targeted. It should in theory be a small subset of the total number of records (hopefully)

WorkSutton:
@go8soft Good call on the backup prior to execution. I should have put that in the PR description at the top.

lib/tasks/historic_data_contacts_whitespace_strip.rake
go8soft: sa`about failures` thanks a good point, thanks! Can we do `save!` instead of `save`?

go8soft (changes requested):
Please log the id before `save` and use `save!` instead of `save`. Thanks!

lib/tasks/historic_data_contacts_whitespace_strip.rake
WorkSutton: @go8soft Actioned as requested

userman123:
Happy with this - I want to test it locally and on staging before I merge in though - given it's not going to cause any conflicts, there is no rush to merge in. Will set aside some time in the new year.

PR #374, opened 2020-12-17
Add the DOB to the name for duplicates checking

WorkSutton:
Include DOB as part of rejection validation criteria

[Trello 91](https://trello.com/c/5C483seG/91-include-dob-in-name-check-for-data-import)


userman123 (approved):
looks good I think

PR #377, opened 2020-12-17
Add support for hotreload in docker

dtt101:
Including removing an unneeded step in setup.

Uses foreman to run the app, includes livereload config, and docker specific config for development


Dockerfile.dev
```diff
@@ -2,7 +2,7 @@ FROM ruby:2.5.8
 
 RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
 RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
-RUN apt-get update -qq && apt-get remove cmdtest && apt-get install -y nodejs postgresql-client yarn
```
dtt101, line 5: `cmdtest` is not present on the image so we don't need to remove it

config/environments/development.rb
```diff
@@ -1,6 +1,15 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
   
+  # docker specific development setup
```
dtt101, line 4: Docker specific, including discovering the docker ip so that web console works

Gemfile
```diff
@@ -29,6 +29,11 @@ gem 'activerecord-import'
 
 group :development, :test do
   gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
+  gem 'guard', '~>2.16.2', require: false
+  gem 'guard-livereload', '~>2.5.2', require: false
+  gem 'guard-minitest', '~>2.4.6', require: false
+  gem 'rack-livereload'
+  gem 'foreman'
```
WorkSutton, line 36: @dtt101 Does foreman work in Docker?  I have never tried it as projects recently that have been using docker simply start up multiple containers (services). e.g webpacker is normally its own service in the docker-compose file but it runs the same image as the main app

WorkSutton: @dtt101 Do we not get hot-reloading with webpacker? I know all of our assets are not currently in packs, but it seems to be more common to moving more to be under the control of webpacker. And effectively removing everything out of the asset pipeline.

Gemfile
```diff
@@ -29,6 +29,11 @@ gem 'activerecord-import'
 
 group :development, :test do
   gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
+  gem 'guard', '~>2.16.2', require: false
+  gem 'guard-livereload', '~>2.5.2', require: false
+  gem 'guard-minitest', '~>2.4.6', require: false
```
WorkSutton, line 34: @dtt101 I can see we have minitest in this project, but we are not actively using it as far as I can see. Having a quick scan of the documentation for the gem, it seems to be only for minitest. Would we not be better using the `rspec` focused version? And one for `cucumber` too if it exists?

My suspicion is `minitest` is in our Gemfile lock as a result of how the project was created. At creation time with `rails new` that the option to drop test was not specified.

config/environments/development.rb
```diff
@@ -1,6 +1,15 @@
...
+  if ENV.fetch('DOCKERIZED', false) == 'true'
+    # setup hot reload
+    config.middleware.insert_after(ActionDispatch::Static, Rack::LiveReload)
+    # whitelist ip to enable web console
+    host_ip = `/sbin/ip route|awk '/default/ { print $3 }'`.strip
```
WorkSutton, line 9: Very nice. Its not every day one sees `awk` used. And especially not from within ruby! Very elegant solution.

In the past I have seen this setup web console simply plugged in as a entry where you have deleted it from.
This is a far better solution as its dynamic. Super nice @dtt101

WorkSutton (commented):
I am wondering why we want to use a Guardfile for hot-reloading. 

I used to use it years ago pre Docker and webpacker days. But it seems to have fallen out of fashion (or maybe enough devs simply don't know about it - or rely on their CI systems to quickly run the tests).

No issues with the actual code, more I had some questions on reasoning.

WorkSutton:
Another question for another ticket @dtt101 - given we have an alpine variant for postgres, what are your views on having an app based on Alpine vs the current Debian?

Dockerfile.dev
dmgarland: But we do nodejs postgresql-client and yarn...

Gemfile
dtt101: Good catch! I will remove the `guard-minitest` gem and leave hot reload of tests for another PR, just to keep this one simple.

Gemfile
dtt101: Good questions (on guard, webpacker etc.)
I'll pick them all up here as they are kind of related.

1. I have used webpacker for react/js hot reloading but not for rails 'backend' code before, so part of the choice for Guard is that I have configured it before and know how it works. This may or may not be a good thing!
2. Previous to this PR the app currently, when run in docker, doesn't support any kind of hot reload (i.e when I change a html.erb file, I need to Ctrl + R to see the changes in the browser)
3. foreman does work in docker - used here because it felt like a nice way of encapsulating the commands

config/environments/development.rb
dtt101: Totally google searched it, but also thought it was a nice solution! So good to get the review :)

dtt101:
@WorkSutton - for alpine definitely a fan - the less things to worry about in the image the better! I have only used node alpine before so can't comment specifically on the postgres one!

Gemfile
dtt101: I should add - totally happy to use webpacker if it is the thing!

WorkSutton: @dtt101 Thanks for the great reply. The explanation is sound. 

I was more curious as to why we might want guard. Like you I have more experience with it than the newer Rails features. 

I find it confusing that we - Rails developers - have webpacker but only want it for js and not all assets. But thats probably because most recent experiences have been backend focused and mainly API's. So all the FE stuff is in webpacker or whatever the trend is for JS/CSS management is (bower/webpack/etc).

The Asset pipeline was I think probably well ahead of its time.

Great to know about being able to use `foreman` in docker.

I'm trying to learn more about webpacker - so I can make the right decisions around if everything should move to it or not. Its still actively discouraged from the reading I do. Having said that not everybody agrees with DHH's or the Rails core teams thinking.

I personally like live-reload via Guard, but it did seem to me to go out of fashion.

WorkSutton:
> @WorkSutton - for alpine definitely a fan - the less things to worry about in the image the better! I have only used node alpine before so can't comment specifically on another one!

Sometimes there can be issues on the dev side due to missing dependencies, or not like for like - but generally from what I read - Alpine is pretty well supported.


Having said that, I read a really interesting blog about a month or so back that gave a compelling reason for using Debian for Dev due to some incompatibility issues.

For prod I think Alpine is definitely the way to go, and also to use multi-stage builds (which I am still learning about).

WorkSutton (approved):
Great explanation and discussion @dtt101 . Thank you. Giving it my approval

Gemfile
dtt101: That's super interesting - thanks for sharing the thought process, that's super helpful.
If we want to merge this now, just to get the feature in, I am definitely happy to swap it out for webpacker down the line?

What do you reckon? It is all quite discreet and not much code.

dtt101:
@WorkSutton - thank you for the review! Much appreciated!

PR #378, opened 2020-12-17
Change the start date of a need to be todays date

WorkSutton:
[Trello 110](https://trello.com/c/X5K1YiQW/110-import-starts-on-date-needs-to-be-the-day-of-import-rather-than-2-days-time)

Used this as test data

[20201216-123600-test-and-trace.xlsx](https://github.com/wearefuturegov/beacon/files/5709666/20201216-123600-test-and-trace.xlsx)

Row 7 - Natalia is used for an import with the old calculation of `today + 2.days`

110-Old-import-starts-plus-2-days

Row 8 - Dave is used for the import where the Need `start_date = today`
110-Updated-import-starts-today

You can also see that the source for these has changed to reflect discussion in todays meeting of the Note.name becoming `National Test and Trace Outbound Welfare Call`

### DIFF
--- a/app/models/imported_item.rb
+++ b/app/models/imported_item.rb
@@ -43,7 +43,7 @@ class ImportedItem < ApplicationRecord
         telephone: row[8],
         address: row[9],
         postcode: row[10],
-        needs: [Need.new(category: 'Check in', start_on: Date.today + 2.days, name: row[11])],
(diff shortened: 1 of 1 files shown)


userman123:
So this means that the need/check-in no longer appears in 'Future tab' need to check if this is ok

userman123 (approved):

PR #386, opened 2020-12-22
Feature/rejected contacts

dmgarland:
Displays rejected contacts on imported items under a drop down filter

### TODO
- [ ] Remove some fields and show the reason instead


WorkSutton (approved):
I would personally prefer to see only 2 spaces instead of 4 for indentation, but thats because of what I am used to.

Otherwise LGTM

PR #390, opened 2021-01-05
Trello-102: Staging Banner

WorkSutton:
Does what it says on the tin! But not quite as originally requested. 

Delivered: Add-staging-banner-#2

see [102-staging-banner](https://trello.com/c/bxve5vZq/102-staging-banner)


userman123 (approved):
should probs be putting this in style guide but seeing as its ready and useful soon, lets go with it as is

PR #397, opened 2021-01-07
Use env hostname to determine env

userman123:
Bit messy but easiest solution I could think of for now. We couldnt use Rails.env as this is procutoin on staging environment. Would rather not change that as think its good for staging to be as close to prod as possible

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -60,9 +60,9 @@ module ApplicationHelper
   end
 
   def environment_from_hostname
-    return 'staging' if ENV['HOSTNAME'].include?('staging')
-    return 'test' if ENV['HOSTNAME'].include?('test')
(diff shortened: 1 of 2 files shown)


WorkSutton (approved):
Maybe change `No real data` to be `Personally Identifiable Data (PID)` or it might be PII

app/helpers/application_helper.rb
```diff
@@ -58,4 +58,11 @@ def sort_icon(field)
   def sort_direction(_field)
     (SORT_DIRECTIONS - [params[:order_dir]]).first
   end
+
+  def environment_from_hostname
+    return 'staging' if ENV['HOSTNAME'].include?('staging')
+    return 'test' if ENV['HOSTNAME'].include?('test')
+    return 'development' if ENV['HOSTNAME'].include?('localhost')
```
WorkSutton, line 65: Does localhost work in docker?

PR #399, opened 2021-01-08
Trello-138: Reduce number of columns displayed in Import

WorkSutton:
Iceboxing - [](https://trello.com/c/g6cSBOEB/130-import-category-column-not-being-populated)

[This ticket](https://trello.com/c/7am5uO1b/138-too-many-columns)

Existing-add-person
Contact-with-business-name
Old-form-with-addition-of-business-name
Trello-138-too-many-columns-on-import-record

Also combined with [129-reason-for-failure-not-visible](https://trello.com/c/xQuDMaCr/129-reason-for-failure-not-visible)

**Before**:
138-too-many-columns-PRE-changes

Trello-138-and-129-reason-for-failure-BEFORE

**After**:
Trello-138-and-129-reason-for-failure-AFTER

**Test file**:
[20210112-084100-test-and-trace.xlsx](https://github.com/wearefuturegov/beacon/files/5801117/20210112-084100-test-and-trace.xlsx)

### DIFF
--- a/app/views/contacts/_imported_item_contacts.html.erb
+++ b/app/views/contacts/_imported_item_contacts.html.erb
@@ -91,10 +91,18 @@
     </div>
     <div class="o-layout__item u-1-of-10">
       <div class="b-sm__t b-gray u-pb u-pt">
-        <% if contact&.needs&.first&.name %>
-          <%= contact.needs.first.name %>
+        <% if params[:view] == 'Failed' %>
+          <% if contact.needs.blank? %>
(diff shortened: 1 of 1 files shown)


userman123:
A few issues with this:
- table not filling full width of page
- First name dropping below icon if too long
- address taking up multiple lines. before it was just taking up one

To be fair, I think only the first of the above bullet points is to do with this PR

Screenshot 2021-01-08 at 15 22 41

PR #400, opened 2021-01-08
Revert "Trello-126: - Allow ordering of needs based on 'assigned to'"

userman123:
Reverts wearefuturegov/beacon#393


userman123:
Reverting as some issues to be resolved

PR #401, opened 2021-01-08
Trello-131: default sort order is by created_at DESC, N/A when no results

WorkSutton:
[Default sort ordered by most recent imports](https://trello.com/c/gKMTgfCZ/131-sort-by-most-recent-first-post-import)

Trello-131-new-default-sort-order

Also added 'N/A' where no data available for `Imported`, `Rejected` and `Imported By`


userman123:
The issue was after creating a new import and this is still broken
Screenshot 2021-01-11 at 12 32 54

userman123:
fixed
Screenshot 2021-01-11 at 12 41 49

WorkSutton (commented):
LGTM

PR #405, opened 2021-01-11
Undo changing of search path

userman123:
This undoes some of https://github.com/wearefuturegov/beacon/pull/386 as had broken the search

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -7,20 +7,20 @@ class ContactsController < ApplicationController
   helper_method :name_for_lead_service
 
   def index
-    @params = params.permit(:search, :page, :imported_item_id)
-    if params[:imported_item_id] # Viewing contacts for an import
-      model = params[:view] == 'Failed' ? RejectedContact : Contact
-      @contacts = policy_scope(model)
-      @contacts = @contacts.where(imported_item_id: @params[:imported_item_id]) if @imported_item
-      Rails.logger.unknown("User viewed contact list page for imported item id: #{params[:imported_item_id]} for contact ID: #{@contacts.map(&:id).join(',')}")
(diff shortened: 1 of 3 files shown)


app/controllers/contacts_controller.rb
```diff
@@ -7,20 +7,20 @@ class ContactsController < ApplicationController
...
-  end
-
-  def search
     @params = params.permit(:search, :page, :imported_item_id)
-    @contacts = policy_scope(Contact)
-    @contacts = Contact.search(@params[:search]).where(id: @contacts.select(:id))
+    if params[:imported_item_id] # Viewing contacts for an import
+      model = params[:view] == 'Failed' ? RejectedContact : Contact
+      @contacts = policy_scope(model)
+      @contacts = @contacts.where(imported_item_id: @params[:imported_item_id]) if @imported_item
```
WorkSutton, line 14: Could you not use this outside of the if block and still have the model determined by inspecting `params[:view]`?

That way you could remove line 17 I think

app/controllers/contacts_controller.rb
```diff
@@ -7,20 +7,20 @@ class ContactsController < ApplicationController
...
+      Rails.logger.unknown("User viewed contact list page for imported item id: #{params[:imported_item_id} for contact ID: #{@contacts.map(&:id).join(',')}")
+    else  # Viewing contacts on 'People' screen
+      @contacts = policy_scope(Contact)
+      @params = params.permit(:search, :page, :imported_item_id)
+      @contacts = @contacts.where(imported_item_id: @params[:imported_item_id]) if @params[:imported_item_id].present?
```
WorkSutton, line 19: Why is this different from line 14 above? Surely both need to be the same?

userman123:
clsoing and taking a different approach

PR #407, opened 2021-01-11
Trello-141: WIP on adding all Camden Style Guide source code into Beacon

WorkSutton:


lib/assets/stylesheets/_colours.scss
```diff
@@ -0,0 +1,42 @@
...
+  off-white: #F6F6F6
+);
+
+$primary-colour: palette(blue) !default;
+$secondary-colour: palette(green) !default;
+$tertiary-colour: palette(gray) !default;
+
+$palette: map.set($palette, primary, $primary-colour);
+$palette: map.set($palette, secondary, $secondary-colour);
+$palette: map.set($palette, tertiary, $tertiary-colour);
```
WorkSutton, line 28: Anything to do with `sass:map`s are not supported in the version of sass that is bundled with rails (libsass). So we have to find an alternative way to write these, either `long-hand` or maybe using sass lists?

WorkSutton:
#TODO Need to include most of the actual InuitCSS files that are referenced. Only copied across the files from the Camden Style Guide that were in the front-end section.

PR #411, opened 2021-04-23
Bump up rails version to resolve mimemagic issue

futuregov:


futuregov:
not pulling in latest commits from branch for some reason

PR #417, opened 2021-05-17
Uses default time formatter for visual DOB

kittysquee:
Closes [Trello 164](https://trello.com/c/trEZv0pz/164-small-improvement-change-the-date-of-birth-day-month-year-to-be-dd-mm-yyyy-throughout)

I just used the default time formatter in the config folder (`config/initializers/time_formats.rb`)

### DIFF
--- a/app/helpers/contacts_helper.rb
+++ b/app/helpers/contacts_helper.rb
@@ -4,4 +4,8 @@ module ContactsHelper
   def boolean_formatter(boolean)
     boolean ? 'Yes' : 'No'
   end
+
+  def short_date_formatter(date)
+    date&.to_s(:default)
+  end
 end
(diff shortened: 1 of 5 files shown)


tosbourn (approved):
Looks good!

If this project ever adopting a decorator pattern this would be a perfect place to throw the logic. As it stands this fits with the rest of the codebase well.

em-cd (commented):
Agree with @tosbourn, perhaps in this case it would be worth popping the formatting into a helper method? I spotted that there's already a `ContactsHelper` module, it could live in there. That way if we ever need to change how the DOB is formatted, we'll only need to change it in one place.

kittysquee:
@em-cd I've created a method which calls the date formatter as suggested.

em-cd (approved):
Perfect! Thanks 😄

PR #421, opened 2021-05-20
Prepends 0 to contact phone numbers when not there

kittysquee:
Closes [Trello issue #166](https://trello.com/c/kqHPPAOS/166-small-improvement-0-on-the-start-of-phone-numbers)

If a telephone or mobile number does not have a zero or plus at the
start it will have a zero prepended. All existing numbers in the
database are checked and updated with a zero at the beginning if none
exists.


tosbourn (approved):
☎️ looks good!

PR #428, opened 2021-05-26
Adds test for blank phone number field

kittysquee:


em-cd (approved):
Perfect, thanks @kittysquee! 😄

PR #432, opened 2021-05-27
Adds 'Test & Trace Support Payment' as notes category

tosbourn:
Also standardises casing in menu


kittysquee (approved):
Good to go once the Cucumber test is fixed

PR #439, opened 2021-06-15
Upgrade Ruby version

userman123:


Dockerfile.dev
```diff
@@ -1,8 +1,12 @@
-FROM ruby:2.5.8
+FROM ruby:2.6.0
+
+ENV BUNDLER_VERSION=2.1.4
 
 RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
-RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+RUN echo "deb http://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
```
em-cd, line 6: Did this really break Docker? 🤯

userman123: yeah..weird I know, I tried installing apt-transport-https ca-certificates as described here, but didn't work so tried their other suggestion of just using http and it solved it 🤷‍♂️  https://unix.stackexchange.com/questions/263801/apt-get-fails-the-method-driver-usr-lib-apt-methods-https-could-not-be-found

em-cd (approved):
LGTM! I got this running locally and did some smoke tests, all looks good 💯

PR #451, opened 2022-01-07
Add created at time to notes

dtt101:
This PR adds a timestamp to notes created 

At some point in the past the app was migrated from Heroku to Cloudfoundry, but the integration tests still run on heroku.
The integration test environment is now failing to buils as heroku has changed the node version, resulting in some of the dependencies of this project being incompatible - a ticket has been raised on the backlog to repair the integration test environment.

PR #452, opened 2022-01-12
Bug fix: order notes by created_at not ID

dtt101:
ID assumes sequential ordering in the DB, which is not guaranteed

As with #451 - the integration test runners on heroku are currently out of order:

tests passing evidence: 
image


em-cd (approved):
LGTM! :shipit:

PR #453, opened 2022-01-12
Postcode search ignores space

dtt101:
Enabling users to search regardless of space characters in strings


dtt101:
As before - remote running of integration tests is borked on heroku:

image

em-cd (approved):
Looks good! thanks @dtt101 🙌

